PR #1, opened 2022-05-10
First commit - move app services

cbush:
https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/move-app-services/


nathan-contino-mongo (approved):
Let's be honest, I'm not going to look at 50,000 files. In fact, my laptop will barely let me type this message since the GitHub UI is chugging so much on this diff. LGTM.

PR #30, opened 2022-05-27
Fix redirects file prefix/base

cbush:
This was not set up correctly


biniona-mongodb (approved):

config/redirects
```diff
@@ -1,8 +1,8 @@
 define: prefix docs/atlas/app-services
-define: base https://www.mongodb.com/docs/
+define: base https://www.mongodb.com/${prefix}
```
cbush, line 2: cf https://github.com/mongodb/docs-realm/blob/master/config/redirects#L2

PR #43, opened 2022-06-06
Move CLI login info to be with CLI stuff

cbush:
https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/move-cli-login/cli/


source/cli.txt
```diff
@@ -39,6 +39,22 @@ Installation
 
 .. include:: /includes/install-realm-cli-v2.rst
 
+Authentication
```
dacharyc, line 42: Hmm, I think I like this on the CLI page - not so much as a separate page within the CLI docs. Can we kill the `cli/auth-with-api-token` page and just redirect to here? (I wonder if there's a reason we had it separate before, like maybe the TSEs link to it - or maybe it's just a leftover from before we had generated CLI docs/better CLI docs.)

dacharyc (approved):
Maybe we can eliminate the separate page entirely and just let it live on `source/cli`?

source/includes/steps-realm-cli-generate-api-key.yaml
```diff
@@ -22,12 +22,17 @@ content: |
      :alt: Click Access Manager
      :lightbox:
 
-  Clicking this button will navigate you to the "Create API Key" screen.
-  Set a description for your key and a project permission. The various
-  project permissions are described on the right hand side of the screen
-  and can be changed at a later time. Copy the public key to use later
-  in order to log in. Click :guilabel:`next` to continue configuring your
-  key details.
+  Clicking this button will navigate you to the "Create API Key" screen. Set a
```
nathan-contino-mongo, line 25:
```suggestion
  Clicking this button navigates you to the "Create API Key" screen. Set a
```
Passive -> active voice, take it or leave it.

source/includes/steps-realm-cli-generate-api-key.yaml
```diff
@@ -22,12 +22,17 @@ content: |
...
+  description for your key.
+
+  For write access, the CLI requires an API key with Project Owner-level
```
nathan-contino-mongo, line 28:
```suggestion
  For write access, the CLI requires an API key with Project Owner
```
Does the "-level" add anything here? if not, let's remove it.

source/includes/steps-realm-cli-generate-api-key.yaml
```diff
@@ -22,12 +22,17 @@ content: |
...
+  permissions. Use the :guilabel:`Project Permissions` dropdown to select
+  "Project Owner".
+
+  If you only need read access, you can use "Project Read Only".
```
nathan-contino-mongo, line 32: It feels a little strange to frontload write permissions here, describe how to select the write permission, then mention the read-only permission as an afterthought. How about:

```suggestion
  For write access, the CLI requires an API key with Project Owner-level
  permissions. If you only need read access, you can use "Project Read Only".
  Use the :guilabel:`Project Permissions` dropdown to select a permission level.
```

nathan-contino-mongo (approved):
LGTM -- some suggestions and thoughts, but nothing blocking.

source/includes/steps-realm-cli-generate-api-key.yaml
cbush: I'll take it, but nit: both are active voice. The thing to avoid here is future tense, which muddies the sentence with modifying verbs.

nathan-contino-mongo: Good point, not passive voice. Glad we're cleaning up the sentence.

PR #49, opened 2022-06-06
Hide flex sync permissions template apps info - temporary

cbush:
These apps will not be available in release.
Revert this change when ready to add.


dacharyc (approved):
LGTM

PR #73, opened 2022-06-27
(DOCSP-23413): Field-level Permissions

cbush:
### Jira

- https://jira.mongodb.org/browse/DOCSP-23413

- https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/field-level-perms/sync/data-access-patterns/permissions/#flexible-sync-session-roles-and-rules

### Review Guidelines

### DIFF
--- a/source/sync/data-access-patterns/permissions.txt
+++ b/source/sync/data-access-patterns/permissions.txt
@@ -408,6 +408,7 @@ A **Role** object has the following properties:
 
 .. list-table::
    :header-rows: 1
+   :widths: 30 20 10 40
 
    * - Key
      - Value Type
@@ -472,6 +473,7 @@ fields object are names of the field:
 
 .. list-table::
    :header-rows: 1
+   :widths: 30 20 10 40
 
    * - Key
      - Value Type
@@ -487,6 +489,7 @@ The **Permissions** object has the following properties:
 
 .. list-table::
    :header-rows: 1
+   :widths: 30 20 10 40
 
    * - Key
      - Value Type
--- a/source/sync/data-access-patterns/permissions.txt
+++ b/source/sync/data-access-patterns/permissions.txt
@@ -308,13 +308,6 @@ Flexible Sync Session Roles and Rules
 In flexible sync, a **session role** determines which permissions apply for the
 duration of a sync session.
 
-A session role consists of:
-
-- A name 
-- An "apply when" :ref:`rule expression <rule-expressions>`
-- A set of :ref:`read <flexible-sync-read-permissions>` and :ref:`write 
-  <flexible-sync-write-permissions>` rule expressions
-
 You define Flexible Sync session roles on a per-collection basis. Each
 "collection" when using {+sync+} corresponds to a Realm Object type.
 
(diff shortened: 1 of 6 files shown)


source/sync/data-access-patterns/permissions.txt
```diff
@@ -360,6 +353,183 @@ change until the next session.
...
    - You cannot store permissions information (such as "which document IDs may this user access?") in the user object.
 
+
+.. _flexible-sync-roles-json-reference:
+
+Flexible Sync Roles JSON Reference
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A **Role** object can specify a specific collection's role or the default role
+used for any unspecified collection. You define field-level permissions with a
```
MongoCaleb, line 363: What do you mean by "any unspecified collection"? It implies, I think, that the default role doesn't apply to a specified collection...which I know isn't sensible but might not be clear to the reader. Would this be more clear:
"or to the default role applied to all collections"?

source/sync/data-access-patterns/permissions.txt
```diff
@@ -360,6 +353,183 @@ change until the next session.
...
+
+   The following JSON object describes a **Role** for a team admin. When granted this role, the user:
+
+   - May access any document for reading.
+   - May access a document for writing if the document's ``teamId`` field equals
+     the user's custom data ``teamId``.
+   - May write to any field of the ``address`` embedded object except
+     ``zipCode``.
+   - May write to the ``name`` field.
+   - May not read or write any other fields despite having access to the
```
MongoCaleb, line 376: This is the only one in which you list the _reason_ for the resulting permissions ("due to the ``additional_fields``...") 
I would prefer to see this in each of the bullet points. Maybe a table?
Role Permissions | Reason
---------------------------
Read any document | "read": {}
---------------------|----------
etc

source/sync/data-access-patterns/permissions.txt
```diff
@@ -489,3 +653,26 @@ A user with write permissions for a given collection can modify the value of any
...
+.. _flexible-sync-field-level-permissions:
+
+Field-Level Permissions
+^^^^^^^^^^^^^^^^^^^^^^^
+
+You can refine document-level permissions on a per-field basis with the
+``fields`` and ``additional_fields`` properties of the **Role** configuration
+object.
+
+If the user does not have document-level permissions to a document, field-level
```
MongoCaleb, line 666: Is this worth putting in a note to make it stand out a bit more? It feels like a "gotcha"

source/sync/data-access-patterns/permissions.txt
```diff
@@ -489,3 +653,26 @@ A user with write permissions for a given collection can modify the value of any
...
+permissions cannot grant access to that document. Field-level permissions can
+only refine the permissions on documents that a user can access. For example,
+you can use field-level permissions to forbid writes on a specific field of a
+document that a user can otherwise write to.
+
+``additional_fields`` defines the field-level permissions for fields that you
```
MongoCaleb, line 672: Hate to see a sentence start with lowercase. How about "The ``additional_fields``property defines..."

source/sync/data-access-patterns/permissions.txt
```diff
@@ -489,3 +653,26 @@ A user with write permissions for a given collection can modify the value of any
...
+don't explicitly define in the ``fields`` object. If you don't define
+``additional_fields`` read or write, the system uses the corresponding
```
MongoCaleb, line 674: "``additional_fields`` read or write ? Do you just mean the "`additional_fields`` property"?

source/sync/data-access-patterns/permissions.txt
```diff
@@ -489,3 +653,26 @@ A user with write permissions for a given collection can modify the value of any
...
+document-level permission for any field not mentioned in ``fields``.
+
+In practice, you would use ``additional_fields`` with ``fields`` to greatly
```
MongoCaleb, line 677: This reads a bit like you're saying "If you know what you're doing, you'll use both properties". Also, "greatly constrain" is perhaps better as "constrain to a higher degree" or "constrain to a finer level of detail"

source/sync/data-access-patterns/permissions.txt
```diff
@@ -360,6 +353,183 @@ change until the next session.
...
+A **Role** object can specify a specific collection's role or the default role
+used for any unspecified collection. You define field-level permissions with a
+**Fields** object in the Role.
+
+.. example::
+
+   The following JSON object describes a **Role** for a team admin. When granted this role, the user:
+
+   - May access any document for reading.
+   - May access a document for writing if the document's ``teamId`` field equals
```
jonatlee, line 371: I know it's mentioned under the "write" Notes, but wondering if it would be worth reiterating that ``teamId`` has to be a queryable field here

source/sync/data-access-patterns/permissions.txt
```diff
@@ -360,6 +353,183 @@ change until the next session.
...
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 10 50
+
+   * - Key
+     - Value Type
+     - Required
+     - Notes
+
+   * - write
```
jonatlee, line 499: Do you think it would be useful to also say that undefined "write" and "read" by default mean that the field has access?

jonatlee (approved):
Looks awesome! Just two small questions/comments

source/sync/data-access-patterns/permissions.txt
cbush: Ah yes. Nice. But...

# Why are you yelling!

😆

source/sync/data-access-patterns/permissions.txt
cbush: Tried to clarify. Take a look.

source/sync/data-access-patterns/permissions.txt
cbush: I get how it works in that we are assumed to have access by the time we even look at the field-level permissions, but I'm concerned about how to communicate that practically. What if we say "if undefined, it grants the document-level permission for this field" or something like that?

source/sync/data-access-patterns/permissions.txt
```diff
@@ -360,6 +353,222 @@ change until the next session.
...
+
+      * - Role JSON
+        - Effect
+
+      * - ``"read": {}``
+        - May access any document for reading.
+
+      * - ``"write": { "teamId": "%%user.custom_data.teamId" }``
+        - May access a document for writing if the document's ``teamId`` field
+          equals the user's custom data ``teamId``. This assumes ``teamId`` was
```
MongoCaleb, line 411: assumes => requires

source/sync/data-access-patterns/permissions.txt
```diff
@@ -360,6 +353,222 @@ change until the next session.
...
+                },
+                "additional_fields": {
+                  "write": true
+                }
+              },
+              "name": {}
+            },
+
+        - May write to any field of the ``address`` embedded object except
+          ``zipCode``. May write to the ``name`` field.
```
MongoCaleb, line 432: May **also** write to the ``name`` field.

source/sync/data-access-patterns/permissions.txt
```diff
@@ -360,6 +353,222 @@ change until the next session.
...
+              "write": false,
+              "read": false
+            }
+
+        - May not read or write any other fields despite having document-level
+          access to the document.
+          
+
+
+A **Role** object has the following properties:
```
MongoCaleb, line 446: I think this whole section might benefit from this coming _before_ the specific example.

MongoCaleb (approved):
LGTM with a couple final suggestions

source/sync/data-access-patterns/permissions.txt
cbush: Cool. I had gone back and forth. I think I agree with you, let's try it!

source/sync/data-access-patterns/permissions.txt
jonatlee: Yeah I think that's fine

PR #89, opened 2022-07-07
(DOCSP-23176): Updated JWT configuration pages to reflect added UI toggle for multiple audiences.

chloe-ratte:
## Pull Request Info

EditED the [JWT Configuration section / Audience subsection](https://www.mongodb.com/docs/atlas/app-services/authentication/custom-jwt/#configuration).  This edit will reflect a change to the optional audience property. Now, when you define multiple audiences in the audience input box, a UI toggle will show up, and offer to require either “All of these audiences” or “Any one of these audiences”.

### Jira

- https://jira.mongodb.org/browse/DOCSP-23176

- [Custom JWT Authentication](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-23176/authentication/custom-jwt/#configuration/)

### Review Guidelines


source/authentication/custom-jwt.txt
```diff
@@ -127,8 +140,29 @@ value, then you can configure the provider to use that value instead.
...
+        * - ``audience``
+
+          - Array of Strings. The audience or audiences expected
+            to be found in your JWT. 
+
+        * - ``requireAnyAudience``
+
+          - Boolean. If ``false``, all of the listed audiences are 
+            required. This is also the value when there is only one audience.
+            If ``true``, any one of these audiences is required.
```
nathan-contino-mongo, line 165:
```suggestion
          - Boolean. If ``false``, valid JWTs must include all of the listed
            audiences. If ``true``, valid JWTs must include at least one of
            the listed audiences.
```
Suggestion: avoid using the word "require" in this explanation, since it's also part of the variable itself and using a different word can help certain users understand the definition more easily. My suggested change uses the phrasing "valid JWTs" to make the end result clearer: JWTs that don't fulfill these criteria are considered invalid.

I suspect that the sentence about "only one audience" is unnecessary -- have you tested one audience with both options? It seems like both would work. I'm curious if there's a reason we suggest "false" when it seems like either option would result in the intended behavior (a one-to-one audience match).

chloe-ratte: I think I worded the only one audience thing weird -- as far as I can tell, the toggle option only appears when you list multiple options. If you only list one option, the toggle doesn't appear, so you can't change the requireAnyAudience variable using the UI. When you list the one option, then look at the CLI, requireAnyAudience is set to false automatically.

Or do you mean I should try editing the CLI code directly to see if I can set requireAnyAudience to true even when there's only one audience?

All this said, it might be a sentence that is best left out anyway -- I don't think it provides much useful information to say that it defaults to false when there's only one audience.

nathan-contino-mongo: I meant the latter -- try changing the config and deploying, and see if it matters at all on the backend. I suspect it doesn't, but... you never know!

Understood that the toggle only appears in the GUI if you add more than one audience. But since the variable is always there (non-optional, afaik) in the config, we should probably include *something* if it has to be set to true or false to work properly. Might be a good idea to ask the engineer who implemented the feature, could be some subtlety there we don't see yet.

nathan-contino-mongo (approved):
LGTM!

PR #102, opened 2022-07-13
(DOCSP-23753): Add missing SDK APIs & custom func to delete users

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-23753

- [Delete App Services User](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-23753/users/delete-or-revoke/)

### Review Guidelines

### DIFF
--- a/source/users/delete-or-revoke.txt
+++ b/source/users/delete-or-revoke.txt
@@ -120,11 +120,14 @@ We would then add the :ref:`Project and Application IDs
 <find-your-project-or-app-id>` to the ``apiUrl``.
 
 We can then :ref:`call this function from the SDK 
-<call-function-from-client-sdks>`, passing in an argument. The example
-function below does not take any arguments, and deletes the user who
-calls the function. However, the SDK expects to provide an argument when you 
-call a function, so you can pass in an arbitrary argument like 
-"Delete this user."
+<call-function-from-client-sdks>`. The example function below does not 
+take any arguments, and deletes the user who calls the function. 
+
+In your function, consider that you may want to optionally delete pending 
+users, which involves a second endpoint:
+
+- :admin-api-endpoint:`Delete users <operation/adminDeleteUser>`
+- :admin-api-endpoint:`Delete pending users <operation/adminDeletePendingUser>`
 
 .. code-block:: javascript
 
--- a/source/users/delete-or-revoke.txt
+++ b/source/users/delete-or-revoke.txt
@@ -123,11 +123,14 @@ We can then :ref:`call this function from the SDK
 <call-function-from-client-sdks>`. The example function below does not 
 take any arguments, and deletes the user who calls the function. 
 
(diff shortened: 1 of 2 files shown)


source/users/delete-or-revoke.txt
```diff
@@ -91,6 +103,66 @@ Manually Delete a User
...
+We can then :ref:`call this function from the SDK 
+<call-function-from-client-sdks>`, passing in the ``user.id``.
+
+.. code-block:: javascript
+
+   const apiUrl = "https://realm.mongodb.com/api/admin/v3.0/groups/{insert-your-groupId-here}/apps/{insert-your-appId-here}";
+
+   // deleteAUser() deletes a single user from the database using the Admin API 
+   // pass in an array with an object with the key userID (representing the app user's user.id)
+   // return value if successful = id of the deleted user
```
nathan-contino-mongo, line 129: This seems... very dangerous. Can we add a check so that a given user can only delete their own account? Otherwise, we're creating a function that anyone can use to destroy anyone else's account. I would probably make this function uncallable from the SDKs, and call it from another function that checks the calling user ID.

Unless I'm missing something here, in which case... fair enough!

nathan-contino-mongo (commented):
See my comment -- I'm concerned we might encourage a very dangerous practice here.

source/users/delete-or-revoke.txt
```diff
@@ -91,6 +103,66 @@ Manually Delete a User
...
+adminApiPublicKey and adminApiPrivateKey <values-and-secrets>`. 
+We would then add the :ref:`Project and Application IDs 
+<find-your-project-or-app-id>` to the ``apiUrl``.
+
+We can then :ref:`call this function from the SDK 
+<call-function-from-client-sdks>`, passing in an argument. The example
+function below does not take any arguments, and deletes the user who
+calls the function. However, the SDK expects to provide an argument when you 
+call a function, so you can pass in an arbitrary argument like 
+"Delete this user."
```
nathan-contino-mongo, line 127: I wonder if we should report this to the SDK folks as a bug -- I feel like you should be able to call functions with no argument. Ought to be an easy fix, I would think?

source/users/delete-or-revoke.txt
```diff
@@ -91,6 +103,66 @@ Manually Delete a User
...
+
+.. code-block:: javascript
+
+   const apiUrl = "https://realm.mongodb.com/api/admin/v3.0/groups/{insert-your-project-id}/apps/{insert-your-app-id}";
+
+   exports = async function(){
+
+   // This function deletes the user who calls it. It gets this user's ID
+   // from the user in the function context. This is safer than accepting 
+   // a passed-in user ID, as the user can never delete any other user's account.
```
nathan-contino-mongo, line 137: 💪

nathan-contino-mongo (approved):
LGTM!

PR #106, opened 2022-07-19
Converted all step files to inline procedures

nathan-contino-mongo:
### Jira

- skunkworks 4 life

- [no steps at all](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/nosteps/)

### Review Guidelines


source/sync/data-model/migrate-schema-partner-collection.txt
```diff
@@ -96,4 +96,132 @@ The *new* schema is the same, except we want the ``_id`` field to be a string:
...
+      schema, you must create a partner collection with a schema containing the
+      required changes. You must ensure that the partner collection has the same
+      data as the original collection so that newer clients can synchronize with
+      older clients.
+
+      The recommended approach to copying the data from your original collection to
+      the new partner collection is to use the :manual:`Aggregation Framework
+      </aggregation/>`.
+
+   .. step:: Create an Aggregation Pipeline
```
dacharyc, line 113: In the currently-live version of the page, this is a header within a step instead of a separate step. I don't think this should be a separate step because it's _how_ we're recommending to Initialize the Partner Collection.

dacharyc (approved):
There are a couple of spots where the new step/procedures deviate from the existing live stuff. I can't tell if you manually changed these up or if there was something wonky about the script. One of them should, I think, not be changed, but I don't mind the other one being updated - just wanted to call it out in case you thought we should do something about it.

Thank you for doing this - nice work!

source/functions.txt
```diff
@@ -213,12 +213,322 @@ UI or by importing the function configuration and source code with
...
+                 "name": "<Function Name>",
+                 "private": <Boolean>,
+                 "can_evaluate": { <JSON Expression> },
+                 "disable_arg_logs": <Boolean>,
+                 "run_as_system": <Boolean>,
+                 "run_as_user_id": "<{+service-short+} User ID>",
+                 "run_as_user_id_script_source": "<Function Source Code>"
+               }
+
+         .. step:: Set User Authentication
```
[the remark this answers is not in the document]
nathan-contino-mongo: I separated them out because I don't think headings are compatible with the procedure/step directives -- I've never seen rST display them that way. Will make this a series of sub-steps instead.

source/sync/data-model/migrate-schema-partner-collection.txt
nathan-contino-mongo: Makes sense. I'll see if I can rework this.

PR #120, opened 2022-08-04
(DOCSP-24182): New AppX Landing Page

nlarew:
## Pull Request Info

[Working Copy Doc](https://docs.google.com/document/d/1aEHckpA9AIg-VWN3apd23ne12AHbn5c_6U9G9uQ3flA/edit#)

### Jira

- (DOCSP-24182): New AppX Landing Page

- [Atlas App Services](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/landing-page/)

### Review Guidelines


dacharyc (commented):
I think this looks great, and I really like it versus the card-style landing page we've currently got. But I think we could make some more tweaks to make it more valuable to devs (and potentially less sales-ey). Curious to hear your thoughts on these comments!

source/index.txt
```diff
@@ -32,145 +38,351 @@
...
-.. button:: Create an Account
-   :uri: https://www.mongodb.com/realm/register?tck=docs_CTA_app_services
+   Services let you write the code and configuration that matters for
+   your app without worrying about how to provision and manage the
+   infrastructure it runs on.
+   
+   .. button:: Start Building
+      :uri: https://www.mongodb.com/realm/register?tck=docs_CTA_app_services
+   
+   .. note:: Usage-Based Pricing & Free Tier
```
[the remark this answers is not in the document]
sumedhamehta: Maybe it's worth putting this at the bottom? I do think people care about the free-tier/ask about it

source/index.txt
```diff
@@ -32,145 +38,351 @@
...
 
-   .. card::
-      :headline: Explore Template Applications
-      :url: https://www.mongodb.com/docs/atlas/app-services/template-apps/
-      :icon: /images/icons/realm/realm-mobile.svg
-      :icon-alt: {+service-short+} Mobile Icon
-      
-      Explore working applications to learn how to implement {+sync+} in 
-      mobile applications or use App Services.
+.. tabs::
```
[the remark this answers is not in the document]
sumedhamehta: I love this idea - I think we could even extend to each service's tutorial at the bottom (or template for sync)

source/index.txt
```diff
@@ -32,145 +38,351 @@
...
-      :headline: Learn More About Services
-      :url: https://www.mongodb.com/docs/atlas/app-services/introduction/
-      :icon: /images/icons/realm/graphql.svg
-      :icon-alt: GraphQL Icon
+   .. tab:: Device Sync
+      :tabid: device-sync
+
+      .. procedure::
+
+         .. step:: Create Your Data Model
```
sumedhamehta, line 78: For all the services/steps, I did a pass on this doc on how to best talk about these services from someone who is coming in from Atlas's POV/making the steps a bit more contextual to the level of knowledge they may have about document database. Let me know what you think https://docs.google.com/document/d/1aEHckpA9AIg-VWN3apd23ne12AHbn5c_6U9G9uQ3flA/edit#

sumedhamehta (changes requested):

source/index.txt
```diff
@@ -32,145 +38,297 @@
...
+            activity as the occur or run triggers on a schedule. You can
+            also set additional conditions for when your trigger can
+            run. For example, you can set a database trigger to run only
+            for events where a document's date property was updated.
+
+         .. step:: Write the Trigger's Handler Logic
 
-Explore {+service+}
---------------------------
+            Define what happens when the trigger runs in response to an
```
sumedhamehta, line 222: This is still on the wordier side, thoughts on 


"Define a serverless JavaScript function to run when the trigger fires or send the Trigger Eventbridge"?

source/index.txt
```diff
@@ -32,145 +38,297 @@
...
+            Once deployed, the trigger automatically listens for events
+            in your app and runs whenever an event matches the trigger’s
+            configuration.
+
+      .. image:: /images/plp/triggers.png
+         :alt: Respond to Events with Triggers
+
+.. kicker:: Power Your App
+
+Run And Secure Your Services With A Fully-Managed Application Server
```
sumedhamehta, line 241: When I saw this on the staged version it read pretty long.. any thoughts on 

"Run your services with additional backend features"

sumedhamehta (changes requested):

source/index.txt
```diff
@@ -32,145 +38,297 @@
    {+client-sdks+} <https://mongodb.com/docs/realm>
    Get Help </help>
 
-{+service+} provides a suite of managed cloud services including {+sync+},
-serverless cloud functions, declarative data access rules, and more. You can use
-{+service-short+} to write and host a full application in a managed cloud
-environment backed by MongoDB Atlas.
+.. introduction::
+   
+   Atlas Application Services are fully-managed backend services and APIs that help you build apps and services with MongoDB Atlas faster. You can use the services to build apps, integrate Atlas to other services, and connect to your Atlas data without managing servers. You can get started with a free tier and check out the usage-based pricing on the :doc:`Billing </billing>` page.
```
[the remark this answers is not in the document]
mongodben: agree. also notice that when using the `:page:` directive it adds a little arrow to the right of `billing`. is this desired?

mongodben:
The image quality for the vector art doesn't look very good on my 24" monitor. are there larger versions of these images that we could be using here?

nlarew:
> also are there 2 separate images we could be using for REST API and GraphQL API? if not, perhaps a separate one for GraphQL could be requested.

@sumedhamehta thoughts?

source/index.txt
```diff
@@ -32,145 +38,329 @@ Atlas App Services
    {+client-sdks+} <https://mongodb.com/docs/realm>
    Get Help </help>
 
-Atlas App Services provides a suite of managed cloud services including Atlas Device Sync,
-serverless cloud functions, declarative data access rules, and more. You can use
-App Services to write and host a full application in a managed cloud
-environment backed by MongoDB Atlas.
+.. introduction::
+   
+   Atlas Application Services are fully-managed backend services and
```
dacharyc, line 43: > .. fully-managed backend services... help you build apps and services... use the services to build apps... other services...

Nit: there is a lot of "services" and "build apps" here. Could we wordsmith this a bit more? Maybe in a follow-up PR because I don't want to block this.

What about something like:

> .. fully-managed backend services and APIs that help you develop faster with MongoDB Atlas. Use App Services to build apps, integrate Atlas...

source/index.txt
```diff
@@ -32,145 +38,329 @@ Atlas App Services
...
+            Device Sync uses Realm's native language objects to define a
+            data model and work with data.
+            
+            To install an SDK, see the :realm:`Realm SDK documentation
+            </>` for your language and platform.
+
+         .. step:: Enable Sync
+
+            In your Atlas App, specify which fields users can query and
+            define read and write permissions in Dev Mode. Once Sync is
```
dacharyc, line 82: Nit: this reads to me like Dev Mode is what enables you to "specify which fields users can query and define read and write permissions." I'm not sure we need to mention Dev Mode at all here, if this is a high-level overview, so I'm wondering if we can just stop at "define read and write permissions."

source/index.txt
```diff
@@ -32,145 +38,329 @@ Atlas App Services
...
+
+   .. tab:: Access Data with a REST API
+      :tabid: data-api
+
+      .. procedure::
+
+         .. step:: Enable the Data API CRUD & Aggregation Endpoints
+
+            The MongoDB Atlas Data API lets you read and write data in
+            Atlas with standard HTTPS requests. Once enabled, you can
```
dacharyc, line 106: Nit: I'm not sure we need "standard" here. This made me stop and think "wait, what do they mean 'standard'?" If we can lose "standard" without losing important meaning, I think we should cut it.

source/index.txt
```diff
@@ -32,145 +38,329 @@ Atlas App Services
...
+         .. step:: Enable Sync
+
+            In your Atlas App, specify which fields users can query and
+            define read and write permissions in Dev Mode. Once Sync is
+            turned on, you can start syncing from any app with the Realm
+            SDKs.
+
+            To enable Sync, see :ref:`Enable Atlas Device Sync <enable-sync>`.
+
+         .. step:: Sync Device Data to Atlas
```
dacharyc, line 88: When I read this tab by itself, this step seemed fine. But after reading some of the other tabs, I realize that all the other steps describe actions that the reader must take, and this step isn't an action. I wonder if we could reword the step title to reflect the action - something like "Read and write data on a device", and then I think the step's text can probably stay as-is or be minimally tweaked.

source/index.txt
```diff
@@ -32,145 +38,329 @@ Atlas App Services
...
-      Learn more about the problems App Services solves - and how - to 
-      decide if App Services fits your use case.
+      .. procedure::
 
-Explore Atlas App Services
---------------------------
+         .. step:: Create and configure your Trigger
+
+            You can respond to database CRUD events and application user
+            activity as the occur or run triggers on a schedule. You can
```
dacharyc, line 251: "as the occur" needs to be fixed, but I'm not sure what the right fix is. "as they occur" would be fine if we're referring to CRUD events, which is plural, but not appropriate for application user activity, which is singular. Maybe something like:

"...database CRUD events and application user activities as they occur..."

source/index.txt
```diff
@@ -32,145 +38,329 @@ Atlas App Services
...
+            fire.
+
+         .. step:: Write the Trigger's Handler Logic
+
+            Define a serverless JavaScript function that runs when the
+            trigger fires. Or, send an event to AWS EventBridge.
+            
+            See :doc:`Trigger Function Examples </triggers/examples>`.
+
+         .. step:: Automatically Respond to Events
```
dacharyc, line 262: Like the Sync tab, this isn't really an action the reader must take but a result of setting up the thing. Maybe we could reword this step title to something like "Run Your App", and then the text still works?

dacharyc (approved):
This is awesome! I've made a handful of wording suggestions, but they're pretty subjective and none of them are blockers. I did spot one grammar fix that needs to be made, though, in the "Build Event-Driven Services" tab.

source/index.txt
nlarew: Tweaked the wording here, removed dev mode specifically

source/index.txt
nlarew: technically we mean [RFC 2818: HTTPS Over TLS](https://www.rfc-editor.org/rfc/rfc2818) but we can probably just omit it here

source/index.txt
nlarew: Good thought - updated the text

PR #127, opened 2022-08-19
Recover custom collection rules

dacharyc:
## Pull Request Info

Based on a Slack request.

- [Sync Rules and Permissions -> Collection-Level Rules](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/recover-collection-based-permissions/sync/data-access-patterns/permissions/#collection-level-rules): Recovered section pertaining to Collection-Level Rules, with some tweaks based on how the page has changed between when we lost the section and now.

### Review Guidelines

### DIFF
--- a/source/sync/data-access-patterns/permissions.txt
+++ b/source/sync/data-access-patterns/permissions.txt
@@ -548,10 +548,10 @@ information, see :ref:`{+sync+} Configuration Files <appconfig-sync>`.
 Custom Rules Per Collection
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You can combine custom rules per collection with default roles. The custom
-session role applies to the ``Employees`` collection. The ``Store`` 
-collection is not listed in the configuration, and has no custom session 
-role, so it also evaluates to the default role.
+You can combine custom rules per collection with default roles. This example 
+shows applying a custom rule to the ``Employees`` collection. This app also 
+links to a ``Store`` collection that is not listed in the configuration, 
+and has no custom collection rule, so it evaluates to the default role.
 
 Sync attempts to find a matching session role by traversing the session
 roles in descending order. List the most specific custom session roles
--- a/source/sync/data-access-patterns/permissions.txt
+++ b/source/sync/data-access-patterns/permissions.txt
@@ -345,21 +345,15 @@ objects regardless of their type. You use :ref:`expressions
 <expressions>` to determine when a role applies and to define dynamic
 field-level permissions for a specific role.
 
-.. _flexible-sync-collection-level-rules:
-
-Collection-Level Rules
-~~~~~~~~~~~~~~~~~~~~~~
-
-You can create rules that apply to an entire collection, and combine
-them with default roles. This example shows applying a custom rule to the 
-``Employees`` collection. This app also links to a ``Store`` collection 
-that is not listed in the configuration, and has no custom collection rule, 
-so it evaluates to the default role.
+You can use type-specific roles alongside default roles. This example shows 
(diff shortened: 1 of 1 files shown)


source/sync/data-access-patterns/permissions.txt
```diff
@@ -345,6 +345,73 @@ objects regardless of their type. You use :ref:`expressions
...
+them with default roles. This example shows applying a custom rule to the 
+``Employees`` collection. This app also links to a ``Store`` collection 
+that is not listed in the configuration, and has no custom collection rule, 
+so it evaluates to the default role.
+
+Sync attempts to find a matching session role by traversing the session
+roles in descending order. List the most specific custom session roles
+first, getting gradually more general, so the user "falls through" to
+the correct session role. If no role applies, and you have defined a
+default role, the user gets the default role permissions.
```
nlarew, line 363: Technically the default role needs to apply to - they all have apply when expressions

source/sync/data-access-patterns/permissions.txt
```diff
@@ -345,6 +345,73 @@ objects regardless of their type. You use :ref:`expressions
...
+
+.. code-block:: json
+
+   {
+    "rules": {
+      "Employees": [
+        {
+          "name": "employeeWriteSelf",
+          "applyWhen": {},
```
nlarew, line 372: nit: not sure if we want to show a situation where you could fall through to default roles - since this always applies that won't happen here

dacharyc: Would you suggest a different example?

source/sync/data-access-patterns/permissions.txt
```diff
@@ -345,6 +345,74 @@ objects regardless of their type. You use :ref:`expressions
 <expressions>` to determine when a role applies and to define dynamic
 field-level permissions for a specific role.
 
+.. _flexible-sync-collection-level-rules:
+
+Collection-Level Rules
```
cbush, line 350: Do we want to introduce the word "collection" here? In a Realm context we can say "Type-specific". There's no obvious connection between object types and "collection". Would it work if this section got merged with the above section?

dacharyc: 🤔 

Originally, this section was titled "Custom and Default Roles" - but the reason for bringing it back is responding to user who is confused about whether you can assign different permissions on a per-collection basis. And we do talk about collections on the "[Choose Your Sync Mode](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/recover-collection-based-permissions/sync/data-access-patterns/sync-mode/#flexible-sync)" page in the example. I think using this heading makes it scannable on the page for folks who might specifically be looking for collection-level permissions.

cbush: Either way, doesn't it leave us with two sections (this and the one above) that basically say the same thing in different words? Still not sure why this was removed exactly, but I know we were trying to avoid mixing and matching Realm and MongoDB terminology.

dacharyc: Hmm, ok. I've updated it a bit to combine it with the section above and use more of the type-specific language. I did leave collections in a couple of places because I think you can infer that the collection is a collection of the specified type in context, without needing to know that a collection is a MongoDB term. What do you think of the updated version?

As a general note, though, collection is a term used all over the App Services UI, including when configuring Rules and Device Sync. It's probably a good idea to use terms in the App Services docs that are consistent with the App Services UI, and if that differs from the Realm term and we're discussing a Realm concept, just clearly state somewhere how the two terms are related.

cbush (approved):

PR #135, opened 2022-08-30
(DOCSP-23950): Document compensating writes

cbush:
### Jira

- https://jira.mongodb.org/browse/DOCSP-23950

- https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/compensating-writes/sync/error-handling/errors/#flexible-sync-errors

### Review Guidelines

### DIFF
--- a/source/sync/data-access-patterns/permissions.txt
+++ b/source/sync/data-access-patterns/permissions.txt
@@ -302,15 +302,15 @@ cannot express solely in JSON expressions.
 .. _flexible-sync-rules-and-permissions:
 .. _flexible-sync-roles:
 
-Session Roles
--------------
+Flexible Sync - Session Roles
+-----------------------------
 
 A **session role** determines which permissions apply for the duration
-of a flexible sync session.
+of a Flexible Sync session.
 
 .. seealso::
 
-   For a guide to setting up flexible sync with common permissions
+   For a guide to setting up Flexible Sync with common permissions
    models, see :ref:`flexible-sync-permissions-guide`.
 
 When a user begins a session by opening a synced realm, session roles
--- a/source/sync/data-access-patterns/sync-mode.txt
+++ b/source/sync/data-access-patterns/sync-mode.txt
@@ -111,6 +111,10 @@ the query. App Services translates those documents to objects in the client
 application, making it easy to work with objects in your app. You can apply
 permissions on both a per-document and per-field basis.
 
+Flexible Sync is intended to be a superset of all functionality in
+Partition-Based Sync. In other words, anything you can do in Partition-Based
+Sync you can also do with Flexible Sync.
+
 In Flexible Sync, you define a set of :ref:`queryable fields 
 <queryable-fields>` that you can query from the client application. 
 You can then subscribe to queries on those fields from the client 
--- a/source/sync/error-handling/errors.txt
+++ b/source/sync/error-handling/errors.txt
@@ -207,8 +207,8 @@ The following errors may occur when your {+app+} uses :ref:`Flexible Sync
          <client-resets>`.
        - When modifying an object in a collection with :ref:`asymmetric sync
          <optimize-asymmetric-sync>` enabled. In this case, the error is
-         non-fatal and does not trigger a client reset.
-
+         non-fatal and does not trigger a client reset. The server skips over
+         the illegal change and does not apply it to the synced MongoDB cluster.
 
 .. _mongodb-translator-errors:
 
--- a/source/sync/error-handling/errors.txt
+++ b/source/sync/error-handling/errors.txt
@@ -148,19 +148,36 @@ The following errors may occur when your {+app+} uses :ref:`Flexible Sync
        The client attempts to reconnect until this process is complete. Then,
        this error resolves and Sync begins working normally.
     
-   * - ErrorWriteNotAllowed
-     - This error occurs when a client attempts a write that is disallowed 
-       by permissions, or if the write modifies an object outside of the 
-       subscription query. 
-       
-       Consider an example where the client application opens a subscription 
-       on a ``Places`` object type with the query ``name = "NY"``. If the 
-       client then tries to insert a ``Places`` object with a field 
-       ``name: "CT"``, the write triggers an ``ErrorWriteNotAllowed`` because 
-       the object is not within the subscription query.
+   * - ErrorCompensatingWrite
+     - This non-fatal error occurs when a client attempts an "illegal" write.
+       The following are considered illegal writes:
 
-       This error triggers a client reset. To recover from this error, 
-       perform a :ref:`client reset <client-resets>`.
+       - Creating an object before opening a subscription
+       - Creating an object that would be outside of the client's query view
+       - Modifying an object that is outside of the client's query view
+       - Creating, deleting, or modifying an object or field that the client
+         does not have write permissions to.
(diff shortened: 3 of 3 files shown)


source/sync/data-access-patterns/permissions.txt
```diff
@@ -302,15 +302,15 @@ cannot express solely in JSON expressions.
 .. _flexible-sync-rules-and-permissions:
 .. _flexible-sync-roles:
 
-Session Roles
--------------
+Flexible Sync - Session Roles
```
cbush, line 305: Wasn't clear from the On This Page that this is now Flex Sync territory

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -111,6 +111,10 @@ the query. App Services translates those documents to objects in the client
 application, making it easy to work with objects in your app. You can apply
 permissions on both a per-document and per-field basis.
 
+Flexible Sync is intended to be a superset of all functionality in
+Partition-Based Sync. In other words, anything you can do in Partition-Based
+Sync you can also do with Flexible Sync.
```
cbush, line 116: Sneaking in Ian's note

source/sync/error-handling/errors.txt
```diff
@@ -148,19 +148,36 @@ The following errors may occur when your {+app+} uses :ref:`Flexible Sync
        The client attempts to reconnect until this process is complete. Then,
        this error resolves and Sync begins working normally.
     
-   * - ErrorWriteNotAllowed
```
kmorkos, line 151: Sorry, I should have been more explicit about this in the ticket, but this error is still being used under the following conditions:

1. A client performs an illegal change that _would_ trigger a compensating write, but their SDK version does not support the non-fatal nature of compensating writes.
    - the behavior would match the current description
2. A client performs a change on an asymmetric table that violates their write permissions, and their SDK version includes [this change](https://github.com/realm/realm-core/pull/5690)
    - the behavior would match the current description, except it will be non-fatal and not trigger a client reset (@benredmond has an open PR for this)

I'm not sure if there's precedence for documenting behavior of old SDK versions. If not then bullet 1 above can be ignored

source/sync/error-handling/errors.txt
```diff
@@ -148,19 +148,36 @@ The following errors may occur when your {+app+} uses :ref:`Flexible Sync
...
-
-       This error triggers a client reset. To recover from this error, 
-       perform a :ref:`client reset <client-resets>`.
+   * - ErrorCompensatingWrite
+     - This non-fatal error occurs when a client attempts an "illegal" write.
+       The following are considered illegal writes:
+
+       - Creating an object before opening a subscription
+       - Creating an object that would be outside of the client's query view
+       - Modifying an object that is outside of the client's query view
```
kmorkos, line 157: can we clarify that "query view" includes both the client's subscriptions, as well as their read permissions?

source/sync/error-handling/errors.txt
kmorkos: ah just noticed the callout about (1) in the compensating write error description

source/sync/error-handling/errors.txt
```diff
@@ -148,19 +148,36 @@ The following errors may occur when your {+app+} uses :ref:`Flexible Sync
...
+         
+       Because the local realm has no concept of "illegal" writes, the write
+       will always succeed locally. Upon sync, the server will notice the
+       illegal write. The server then undoes the change. The undo operation
+       ("compensating write") syncs back to the client so the client's realm no
+       longer has the illegal write. The server also sends this error to let the
+       client know what has happened.
+
+       Note: Any local writes between the illegal write and the compensating write
+       will be lost.
```
kmorkos, line 171: can we clarify that this only applies to local writes on the objects being "compensated"

source/sync/error-handling/errors.txt
```diff
@@ -148,19 +148,36 @@ The following errors may occur when your {+app+} uses :ref:`Flexible Sync
...
+       - Modifying an object that is outside of the client's query view
+       - Creating, deleting, or modifying an object or field that the client
+         does not have write permissions to.
+       - Modifying an object in such a way that client would no longer have
+         write permissions on that object or field after the write.
+         
+       Because the local realm has no concept of "illegal" writes, the write
+       will always succeed locally. Upon sync, the server will notice the
+       illegal write. The server then undoes the change. The undo operation
+       ("compensating write") syncs back to the client so the client's realm no
```
nlarew, line 166:
```suggestion
       , called a "compensating write", syncs back to the client so the client's realm no
```

nlarew (approved):

source/sync/error-handling/errors.txt
cbush: Concerned about mentioning asymmetric tables as this would be the first mention and would probably be more confusing. Is there a specific scenario I can describe where this would happen?

kmorkos: Hm, there's no scenario that doesn't involve asymmetric tables. It's only if you try to create an object on an asymmetric table that you don't have write permissions for.

cbush: I see. So, what's an asymmetric table?

source/sync/error-handling/errors.txt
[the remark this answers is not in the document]
cbush: Added one!

source/sync/error-handling/errors.txt
```diff
@@ -162,13 +164,35 @@ The following errors may occur when your {+app+} uses :ref:`Flexible Sync
...
+          Here, the user does not have permission to write to
+          ``fieldA`` but attempts to write to it anyway -- an *illegal* write.
+          The user then performs two *legal* writes to the same object and
+          another legal write to a different object. Upon receiving the
+          compensating write for the illegal write to ``obj1.fieldA``, the
+          subsequent two legal writes to that object are lost. The end result is
+          that obj1 still exists and its two field values are whatever they were
+          before the attempted modifications. Meanwhile, the deletion of obj2 is
+          unrelated to the illegal write that caused the compensating write, so
+          it stands and obj2 stays deleted.
```
kmorkos, line 195: this is awesome! 🚀

source/sync/error-handling/errors.txt
cbush: Ah! Of course. 💡

source/sync/error-handling/errors.txt
```diff
@@ -148,19 +148,67 @@ The following errors may occur when your {+app+} uses :ref:`Flexible Sync
...
+       In the following cases, you'll see ``ErrorWriteNotAllowed`` instead of
+       ``ErrorCompensatingWrite`` after an "illegal" write:
+
+       - When using older SDK versions that are linked to a realm-core version
+         before realm-core 12.1.0. In this case, the server will not undo the
+         illegal write, and you will need to perform a :ref:`client reset
+         <client-resets>`.
+       - When modifying an object in a collection with :ref:`asymmetric sync
+         <optimize-asymmetric-sync>` enabled. In this case, the error is
+         non-fatal and does not trigger a client reset.
```
kmorkos, line 210: also, the server will skip over this change (it won't get applied to mdb, and it will not be "undone" since that doesn't make sense in the context of asymmetric sync)

i'll defer to your judgment if you think this is obvious and can be left out

kmorkos (approved):

PR #149, opened 2022-09-19
(DOCSP-23902): Global URL small fix on Facebook Authentication page

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-23902

- [Facebook Authentication](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-23902-url-fix/authentication/facebook/)

### Review Guidelines

### DIFF
--- a/source/includes/deployment-region-auth-callback-urls.rst
+++ b/source/includes/deployment-region-auth-callback-urls.rst
@@ -8,7 +8,7 @@
    * - | **Global**
      - .. code-block:: text
 
-          https://stitch.mongodb.com/api/client/v2.0/auth/callback
+          https://realm.mongodb.com/api/client/v2.0/auth/callback 
 
    * - | **Virginia**
        | (``us-east-1``)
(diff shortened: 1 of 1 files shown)


source/includes/deployment-region-auth-callback-urls.rst
```diff
@@ -8,7 +8,7 @@
    * - | **Global**
      - .. code-block:: text
 
-          https://stitch.mongodb.com/api/client/v2.0/auth/callback
+          https://realm.mongodb.com/api/client/v2.0/auth/callback
```
nlarew, line 11: IIRC this didn't work when we initially moved from the Stitch -> Realm brand. Did you confirm that this URL works now?

Also, do we need to update the local URLs in the other rows of this table? They all include "stitch" too.

mongomoe: @nlarew we tested out both URLs and didn't have any issues having either the stitch one or the realm one as the redirect URI, which initially seemed a bit strange to me. However, Lindsey also confirmed with product and they verified realm.mongodb.com as the correct one. See: https://mongodb.slack.com/archives/C41SX3C4D/p1665162980386979

PR #150, opened 2022-09-19
(DOCSP-25355) [Admin API] Full trigger definitions

nlarew:
### Jira

- (DOCSP-25355) [Admin API] Full trigger definitions

- [Admin API > Triggers](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/admin-api-triggers/admin/api/v3/#tag/triggers)

### Review Guidelines

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -324,52 +324,42 @@ paths:
     parameters:
       - "$ref": "#/components/parameters/GroupId"
       - "$ref": "#/components/parameters/AppId"
-    post:
-      tags:
-        - triggers
-      operationId: adminCreateTrigger
-      summary: Create a trigger
-      requestBody:
-        description: The trigger to create.
-        required: true
-        content:
-          application/json:
-            schema:
-              "$ref": "#/components/schemas/TriggerRequest"
-      responses:
-        "201":
-          description: Successfully created.
-          content:
-            application/json:
-              schema:
-                "$ref": "#/components/schemas/TriggerResponse"
-              example:
-                name: myTrigger
-                type: DATABASE
-                function_id: 5ae8a0c88f25b9dc5cabb21f
-                config:
-                  operation_types:
-                    - INSERT
-                    - UPDATE
-                    - DELETE
-                    - REPLACE
-                  database: HR
-                  collection: employees
-                  service_id: 5adeb649b8b998486770ae7c
-                  match: {}
-                  full_document: true
     get:
       tags:
         - triggers
       operationId: adminListTriggers
-      summary: List all triggers
+      summary: Get all triggers.
+      description: Get a list of all [triggers](https://www.mongodb.com/docs/atlas/app-services/triggers/overview/) in your application.
       responses:
         "200":
           description: Successfully listed triggers.
           content:
             application/json:
               schema:
-                "$ref": "#/components/schemas/TriggerResponse"
+                type: array
+                description: A list of all database, authentication, and scheduled trigger defined for the app.
+                items:
+                  "$ref": "#/components/schemas/AnyTrigger"
+    post:
+      tags:
+        - triggers
+      operationId: adminCreateTrigger
+      summary: Create a trigger.
+      description: Create a new [trigger](https://www.mongodb.com/docs/atlas/app-services/triggers/overview/) in your application.
+      requestBody:
+        description: A configuration for a database, scheduled, or authentication trigger.
+        required: true
+        content:
+          application/json:
+            schema:
+              "$ref": "#/components/schemas/Trigger"
+      responses:
+        "201":
+          description: Successfully created.
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Trigger"
   "/groups/{groupId}/apps/{appId}/triggers/{triggerId}":
     parameters:
       - "$ref": "#/components/parameters/GroupId"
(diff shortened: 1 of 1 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -4734,152 +4728,235 @@ components:
...
+          operation_type:
+            - CREATE
+          providers:
+            - api-key
+      allOf:
+        - $ref: "#/components/schemas/BaseTrigger"
+        - properties:
+            type:
+              type: string
+              description: The trigger type. For database triggers, this value is always `"AUTHENTICATION"`.
```
dacharyc, line 4906:
```suggestion
              description: The trigger type. For authentication triggers, this value is always `"AUTHENTICATION"`.
```

source/openapi-admin-v3.yaml
```diff
@@ -4734,152 +4728,235 @@ components:
...
+        type: SCHEDULED
+        function_name: createReport
+        config:
+          schedule: "0 8 * * *"
+      allOf:
+        - $ref: "#/components/schemas/BaseTrigger"
+        - properties:
+            type:
+              type: string
+              description: The trigger type. For database triggers, this value is always `"SCHEDULED"`.
```
dacharyc, line 4944:
```suggestion
              description: The trigger type. For scheduled triggers, this value is always `"SCHEDULED"`.
```

source/openapi-admin-v3.yaml
```diff
@@ -4734,152 +4728,235 @@ components:
...
+        - $ref: "#/components/schemas/DatabaseTrigger"
+        - $ref: "#/components/schemas/AuthenticationTrigger"
+        - $ref: "#/components/schemas/ScheduledTrigger"
+    Trigger:
+      oneOf:
+        - $ref: "#/components/schemas/DatabaseTrigger"
+        - $ref: "#/components/schemas/AuthenticationTrigger"
+        - $ref: "#/components/schemas/ScheduledTrigger"
+      discriminator:
+        propertyName: type
```
dacharyc, line 4742: Is this mapping only possible in a property? It took me a minute to figure out that switching the type here changed the configuration value. The `config` object is collapsed by default, so the only change I saw when selecting a different option in the drop-down was the "this value is always "[FOO]"." changing in the `type` description. If we could provide the mapping drop-down in the "Request Body Schema" description - i.e. `A configuration for a database, scheduled, or authentication trigger.` - I think it might be more intuitive to select it there and understand that the subsequent body was changing, even if you couldn't see the config.

Separately, I was also a little puzzled when I saw that the request sample example had a different mapping drop-down that I had to select independent of selecting it here. Is it possible to have the example body derive from the selection here?

source/openapi-admin-v3.yaml
```diff
@@ -4734,152 +4728,235 @@ components:
...
+                - operation_types
+                - full_document
+                - full_document_before_change
+                - unordered
+                - match
+                - project
+              properties:
+                service_id:
+                  type: string
+                  description: The _id value of a linked MongoDB data source.
```
dacharyc, line 4830: Is it possible to link to the endpoints or docs you can use to get this information? This stood out to me when thinking about how to get "the _id value of a linked MongoDB data source" but that might also apply to other things in this list.

dacharyc (commented):
Love this improvement. Back to you with some Qs.

source/openapi-admin-v3.yaml
nlarew: Yeah I can add that here. Probably a larger effort to do this more generally across the API.

source/openapi-admin-v3.yaml
```diff
@@ -400,19 +391,21 @@ paths:
       tags:
         - triggers
       operationId: adminGetTrigger
-      summary: Retrieve the configuration for a specific trigger
+      summary: Get a trigger.
```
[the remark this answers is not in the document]
nlarew: They list in the order they're defined in the spec. In general we try to have plural endpoints (e.g. `/triggers`) before singular endpoints (e.g. `/triggers/{triggerId}`) and then do GET > POST/PUT/PATCH > DELETE within each endpoint.

These do look a little out of order, will fix!

source/openapi-admin-v3.yaml
dacharyc: Nice - I think this UX makes the expected result a little more apparent. I would love to see the example body be linked, but at least now with the pills it seems a little less like recreating the same action and for some reason, that makes it slightly less annoying to have to do it separately.

source/openapi-admin-v3.yaml
```diff
@@ -4734,152 +4728,232 @@ components:
...
+                - unordered
+                - match
+                - project
+              properties:
+                service_id:
+                  type: string
+                  description: |
+                    The _id value of a linked MongoDB data source.
+
+                    See [Get a Data Source](#/operation/adminGetService).
```
dacharyc, line 4827: This link didn't work for me on staging. I _think_ this is the correct fix?
```suggestion
                    See [Get a Data Source](#operation/adminGetService).
```

dacharyc (approved):
Looks like you need to update this after merging the other PR, and I spotted one broken link, but otherwise LGTM. Nice improvement!

source/openapi-admin-v3.yaml
nlarew: Oops good catch!

PR #160, opened 2022-09-27
(DOCSP-25454) Trigger IA refresh

mongodben:
## Pull Request Info

overview of changes: 
- make overview page into landing page
- consolidate the "usage example" create a X trigger with the rest of the documentation on that trigger
- move the trigger examples to the bottom of those pages
- remove the usage examples section

### Jira

- https://jira.mongodb.org/browse/DOCSP-25454

- [Triggers section](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-25454/triggers)

### Review Guidelines


source/triggers.txt
```diff
@@ -1,22 +1,105 @@
...
    :titlesonly:
    :caption: Triggers
    :hidden:
 
-   About Triggers </triggers/overview>
    Database Triggers </triggers/database-triggers>
    Authentication Triggers </triggers/authentication-triggers>
    Scheduled Triggers </triggers/scheduled-triggers>
-   Usage Examples </triggers/examples>
+   Resume a Suspended Trigger </triggers/resume-a-suspended-trigger>
```
nlarew, line 16: Only DB triggers can get suspended like this, so maybe we move this procedure to the DB triggers page?

source/triggers.txt
```diff
@@ -1,22 +1,105 @@
...
+   Disable a Trigger </triggers/disable>
```
nlarew, line 17: Fine to have this as a separate page for now - eventually it'd be great to have this on a "manage triggers" page or somesuch that collects all the various CRUD/admin tasks.

source/triggers.txt
```diff
@@ -1,22 +1,105 @@
...
+   Send Trigger Events to AWS EventBridge </triggers/aws-eventbridge>
+   Store New Users in MongoDB </triggers/store-new-users-in-mongodb>
```
nlarew, line 19: This is specifically an auth trigger example - maybe we consolidate on that page?

nlarew (approved):
LGTM w/ a few comments and suggestions. I think there's some more page-level changes we can make in the future but this refresh definitely is a great start on making things cleaner

source/triggers.txt
mongodben: actually now that you mention it, this example is very familiar to the example already on that page, except less thorough. i'm deleting the `Store New Users in MongoDB` page.

the Triggers example repo will also have an example of this. don't think we need 3 times.

PR #162, opened 2022-09-27
(DOCSP-25463) [Admin API] App CRUD & Deployment

nlarew:
### Jira

- (DOCSP-25463) [Admin API] App CRUD & Deployment

- [Admin API > Applications](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/admin-api-apps/admin/api/v3/#tag/apps)

### Review Guidelines

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -3966,6 +3966,11 @@ components:
         - "GLOBAL"
         - "LOCAL"
     CloudProviderRegionId:
+      description: |-
+        The ID of a cloud provider-specific deployment region. Some
+        regions are available for global deployment while others are
+        only available for local apps. See [Cloud Deployment
+        Regions](https://www.mongodb.com/docs/atlas/app-services/manage-apps/deploy/deployment-models-and-regions/#cloud-deployment-regions).
       oneOf:
         - $ref: "#/components/schemas/AwsDeploymentRegion"
         - $ref: "#/components/schemas/AzureDeploymentRegion"
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -3858,6 +3858,21 @@ components:
           type: string
         disabled:
           type: string
+    DeploymentLocation:
+      type: string
+      description: |-
+        An identifier of the location where an app server is physically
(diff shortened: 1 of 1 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -3827,72 +3858,152 @@ components:
...
+      description: An application [deployment model](https://www.mongodb.com/docs/atlas/app-services/manage-apps/deploy/deployment-models-and-regions/#deployment-models).
+      enum:
+        - "GLOBAL"
+        - "LOCAL"
+    CloudProviderRegionId:
+      oneOf:
+        - $ref: "#/components/schemas/AwsDeploymentRegion"
+        - $ref: "#/components/schemas/AzureDeploymentRegion"
+        - $ref: "#/components/schemas/GcpDeploymentRegion"
+    AwsDeploymentRegion:
```
dacharyc, line 3973: Oh boy. I guess it makes sense to provide this here, but we also need to remember to update it when we add new regions. So far, I think we've just been adding them to the [Deployment Models & Regions page](https://www.mongodb.com/docs/atlas/app-services/manage-apps/deploy/deployment-models-and-regions/#cloud-deployment-regions).

Looking at that Deployment Regions page, should we somehow specify here which ones are restricted to local only? Or maybe just add something in the description that some are only available for local deployment, and to consult the docs?

It also looks like we try to provide values for this in the [App config reference](https://www.mongodb.com/docs/atlas/app-services/manage-apps/configure/config/app/), but those are also not up-to-date. Should we make a separate ticket to update this there, or do you want to tag that onto this PR?

dacharyc (approved):
Overall, LGTM. Just one note about some deployment regions being available only for local deployment, and whether we want to note that here somehow.

source/openapi-admin-v3.yaml
nlarew: I think we should get into the habit of assuming that any configurable change can/will have corresponding Admin API docs updates.

> some are only available for local deployment, and to consult the docs

I think this makes the most sense

> looks like we try to provide values for this in the [App config reference](https://www.mongodb.com/docs/atlas/app-services/manage-apps/configure/config/app/), but those are also not up-to-date

good catch I'll tag on in this PR

PR #170, opened 2022-10-06
(DOCSP-25747) [Admin API] Get & Redeploy a Deployment

nlarew:
### Jira

- (DOCSP-25747) [Admin API] Get & Redeploy a Deployment

- [Admin API > Deployment](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/admin-redeploy/admin/api/v3/#tag/deploy)

### Review Guidelines

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -2581,6 +2581,9 @@ paths:
       - "$ref": "#/components/parameters/GroupId"
       - "$ref": "#/components/parameters/AppId"
   "/groups/{groupId}/apps/{appId}/deployments":
+    parameters:
+      - "$ref": "#/components/parameters/GroupId"
+      - "$ref": "#/components/parameters/AppId"
     get:
       tags:
         - deploy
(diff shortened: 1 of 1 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -2596,9 +2599,51 @@ paths:
...
+      tags:
+        - deploy
+      operationId: adminRedeployDeployment
+      summary: Redeploy a deployment
+      description: Redeploy a previously deployed deployment.
+      responses:
+        "204":
+          description: Successfully redeployed
+        "404":
+          description: Encountered an error while updating the trigger.
```
dacharyc, line 2635: This error description struck me as weird to be talking about triggers while the endpoint is talking about a deployment. Is this error message an example - i.e. it could fail because of something _in_ the deployment - or is that a bug in the endpoint?

source/openapi-admin-v3.yaml
```diff
@@ -2596,9 +2599,51 @@ paths:
...
     parameters:
       - "$ref": "#/components/parameters/GroupId"
       - "$ref": "#/components/parameters/AppId"
+      - "$ref": "#/components/parameters/DeploymentId"
+    get:
+      tags:
+        - deploy
+      operationId: adminRedeployDeployment
+      summary: Redeploy a deployment
+      description: Redeploy a previously deployed deployment.
```
dacharyc, line 2630: Too many deploys. Could we reword this slightly? Maybe something like:

"Redeploy a previously-deployed version of the App" or something?

source/openapi-admin-v3.yaml
```diff
@@ -2596,9 +2599,51 @@ paths:
...
+      responses:
+        "204":
+          description: Successfully redeployed
+        "404":
+          description: Encountered an error while updating the trigger.
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Error"
+                properties:
```
dacharyc, line 2640: This error information doesn't seem to be rendering on the page.

source/openapi-admin-v3.yaml
```diff
@@ -3681,6 +3726,14 @@ components:
       required: true
       schema:
         type: string
+    DeploymentId:
```
dacharyc, line 3729: I'm not seeing this deploymentId in the draft endpoint documentation. Do we need to add this as a param around ln 2734?

dacharyc (changes requested):
Some of the stuff isn't rendering as I'd expect - I think we need some tweaks.

source/openapi-admin-v3.yaml
nlarew: That endpoint uses `DraftId`, not `DeploymentId`. We use `DeploymentId` in the "Get a Deployment" and "Redeploy a Deployment" endpoints.

source/openapi-admin-v3.yaml
nlarew: Oops - missing an `allOf` here, good catch!

source/openapi-admin-v3.yaml
nlarew: Oops copy pasta error - fixed!

source/openapi-admin-v3.yaml
dacharyc: Ah, I missed "current deployment" in the description. Was thinking you could pass both and get a diff. That'll show me for scanning too quickly!

dacharyc (approved):
LGTM!

PR #191, opened 2022-10-20
(DOCSP-25660): Sync landing page

cbush:
- NEW Sync landing page
- Cleaned up 'learn' section
- Moved technical deep dives to new deep dive section
- Created "production checklist"

### Jira

- https://jira.mongodb.org/browse/DOCSP-25660

- https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/sync-landing/sync/

### Review Guidelines


source/sync.txt
```diff
@@ -8,21 +12,129 @@
...
+      current network status.
+
+   .. card::
+      :headline: Conflict Resolution
+      :icon: /images/icons/Technical_REALM_ConflictResolution3x.png
+      :icon-alt: An icon depicting conflict resolution
+
+      {+sync-short+} handles conflicts for you, so you don't have to write
+      complex custom code to resolve conflicting writes from multiple clients.
+      Read more: :ref:`conflict-resolution`
```
dacharyc, line 76: The styling on this :ref: is weird because of the card formatting. Can we change this to a CTA instead?

```suggestion
      :cta: Read more about Conflict Resolution
      :url: https://www.mongodb.com/atlas-app-services/sync/details/conflict-resolution/#std-label-conflict-resolution
      :icon: /images/icons/Technical_REALM_ConflictResolution3x.png
      :icon-alt: An icon depicting conflict resolution

      {+sync-short+} handles conflicts for you, so you don't have to write
      complex custom code to resolve conflicting writes from multiple clients.
```

source/sync.txt
```diff
@@ -8,21 +12,129 @@
...
+
+Key Features
+------------
+
+.. card-group::
+   :columns: 3
+   :style: compact
+
+   .. card::
+      :headline: Realm in the Front, MongoDB in the Back
```
dacharyc, line 52: As much as I think this line is snappy, I'm not sure it works for this case. When I am reading about dev stuff and I hear "front" I think UI, but here we're talking about the client DB. 

Maybe instead this is something like "Realm in the client, MongoDB in the cloud" ?

source/sync.txt
```diff
@@ -8,21 +12,129 @@
...
+   .. card::
+      :headline: Launch a Template Starter App
+      :cta: Register for a free Atlas account
+      :url: https://www.mongodb.com/realm/register?tck=docs_CTA_app_services
+      :icon: /images/icons/Technical_REALM_iOSAndriodCrossPlatform3x.png
+      :icon-alt: An icon depicting two mobile devices synchronizing data between each other
+
+      The best way to start using {+sync-short+} is with a template starter app.
+      In the {+ui+}, under the :guilabel:`App Services` tab, you can select a
+      template. Atlas will configure a backend instance and give you the
```
dacharyc, line 97: Nit: The MongoDB docs style guide suggests avoiding future tense.

```suggestion
      template. Atlas configures a backend instance and gives you the
```

source/sync.txt
```diff
@@ -8,21 +12,129 @@
...
+      frontend code for a selection of platforms. You can build your app on top
```
dacharyc, line 98: Question: do we actively want to recommend building an app on top of the template? Or is it more of a learning resource so devs can see the "aha" moment faster? I worry that if we tell devs to build on this, they're going to get to breaking schema changes and then have a bad experience with the product. As an alternative, I might suggest something like:

> Explore a working example of Sync to learn how to implement it in your app.

source/sync.txt
```diff
@@ -8,21 +12,129 @@
...
+
+Explore the {+sync-short+} Documentation
+-------------------------------------
+
+.. procedure::
+
+   .. step:: Maintain Data Integrity with Sync Schemas
+
+      Define your application data model with standard schemas that you can use
+      to validate data and generate language-specific classes. Read the
```
dacharyc, line 121: I wonder if we want to mention here that we can derive the schema from the client data model? Or if we think that's too much in-the-weeds for this page?

```suggestion
      Define your application data model with standard schemas that you can use
      to validate data. Generate language-specific classes, or infer the schema 
      from your client data model. Read the
```

source/sync.txt
```diff
@@ -8,21 +12,129 @@
...
+      :ref:`sync-schema-overview` to get oriented.
+
+   .. step:: Protect User Data with Permissions
+
+      {+sync-short+} provides a user-based permissions system that lets you
+      control who can access which data. The
+      :ref:`flexible-sync-permissions-guide` guides you through permissions set
```
dacharyc, line 128: This is a total nit, but I don't love "Guide guides you". Maybe we could say instead something like:

> To learn how to set up permissions for a variety of real use cases, check out the :ref:`flexible-sync-permissions-guide`.

source/sync.txt
```diff
@@ -8,21 +12,129 @@
...
+   Production Checklist </sync/production-checklist>
    Sync Production Load Testing </sync/production-load-testing>
    Event Library </sync/event-library>
+   Technical Details </sync/details>
+
+.. introduction::
+
+   You're developing a mobile app. Your users want their data saved on the cloud
+   and accessible from their other devices. Network access on a mobile device
+   can be hit or miss, so you write data locally on the device first. A
```
dacharyc, line 30: Nit: hit or miss is an idiom that may be unfamiliar to folks from other countries or for whom English is a second language. What about an alternative wording?

```suggestion
   can be intermittent, so you write data locally on the device first. A
```

source/sync.txt
```diff
@@ -8,21 +12,129 @@
...
+   background process then synchronizes the data to the cloud and resolves any
+   conflicting writes.
+ 
+   {+sync+} does all that... and more!
```
dacharyc, line 34: Nit: this is a tempting way to phrase this because it's true, but it is pretty informal/casual compared to our tone throughout the rest of the docs, and also reads as a bit salesy to me. We can probably do without this line entirely.

source/sync/production-checklist.txt
```diff
@@ -0,0 +1,66 @@
...
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+
+Best Practices
+--------------
+
+In addition to the information in this guide, we recommend following these best
```
dacharyc, line 18: Nit: this opening sentence for the page doesn't make sense where this is the only content on the page. 
```suggestion
We recommend following these best
```

source/sync/production-checklist.txt
```diff
@@ -0,0 +1,66 @@
+.. _realm-sync-best-practices:
+.. _sync-production-checklist:
+
+================================
+Device Sync Production Checklist
```
dacharyc, line 5: Since we're calling this a production checklist, I wonder if we want to consolidate the [production load testing page](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/sync-landing/sync/production-load-testing/) onto this page?

dacharyc (commented):
I think this is a great move in the right direction! I've got an assortment of nits, suggestions, and comments for you to consider.

source/sync.txt
cbush: That would rob it of its snappiness. I'm not married to this catchphrase, by the way, but I had to get it out of my system...

"Front" in this case means client (front end), "back" meaning server (back end). I'm not sure I understand (agree with) your specific comment here.

source/sync.txt
cbush: Yes, I think that's too far in the weeds.

source/sync.txt
cbush: Fine, but the above paragraph needs a clincher, otherwise I haven't connected it to the actual topic at hand, just described a scenario.

source/sync/production-checklist.txt
cbush: I was thinking we could move both into a subsection called "Go to Production"?

dacharyc: Works for me!

source/sync.txt
cbush: I think the idea is that you can take it and build whatever you want on top of it. Breaking schema changes should be something people can handle in early development - maybe we need a short guide (i.e. remove the app from the simulator, delete cluster data, restart sync)

source/sync.txt
dacharyc: I think the "built for this purpose" that you've added works, or we could say something like "handles this for you." The former says "we did this thing and this is why" but the latter says "this thing solves this problem for you." It's a subtle nuance, so I don't have a strong opinion.

source/sync.txt
dacharyc: Fair. I spend a lot of time talking to my dev friends about dev stuff, and in that circle, "front" generally means "web UI." Perhaps "front end" for a mobile dev is the client and that is a reasonable inference. I was just thinking the cloud/client language is more descriptive/explicit.

cbush: The reason "front" might refer to "web UI" is because a web UI is a client. Clients are typically considered the front end because it's what the user is actually facing in order to interact with the server (backend).

The phrase "frontend developer" might have a connotation of being someone who specializes in CSS (so, web only), but I think this is just because there's less distinction within the mobile developer space. You would expect a "mobile developer" to be able to UI code.

dacharyc (approved):
Nice improvement!

ianpward:
"realm in the front, MDB in the back" is hilarious - what do we think about MulletDB going forward?

cbush:
YES!

source/sync.txt
```diff
@@ -1,28 +1,175 @@
...
-- :doc:`Define Data Access Patterns </sync/data-access-patterns>`
-- :doc:`Configure Sync </sync/configure-device-sync>`
-- :doc:`Handle Errors </sync/error-handling>`
-- :doc:`Add Sync to a Local-Only App </sync/local-to-sync>`
-- :doc:`Sync Production Load Testing </sync/production-load-testing>`
+   Go to Production with Sync </sync/go-to-production>
+
+.. introduction::
+
+   You're developing a mobile app. Your users want their data saved on the cloud
```
mongodben, line 23:
```suggestion
   You're developing a mobile app. Your users want their data saved in the cloud
```

nit

source/sync.txt
```diff
@@ -1,28 +1,175 @@
...
+      :headline: In Flight Security
+      :icon: /images/icons/Technical_REALM_Offline3x.png
+      :icon-alt: An icon depicting network loss
+
+      {+sync-short+} encrypts all network traffic using Transport layer Security
+      (TLS). Document and field-level permissions determine which data may be
+      synced to the device.
+
+   .. card::
+      :headline: On the Cloud
```
mongodben, line 102:
```suggestion
      :headline: In the Cloud
```

nit

PR #192, opened 2022-10-21
(DOCSP-25876) [Admin API] Trigger Event Processors

nlarew:
### Jira

- (DOCSP-25876) [Admin API] Trigger Event Processors

- [PAGE_NAME](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/BRANCH_NAME/)

### Review Guidelines


source/openapi-admin-v3.yaml
```diff
@@ -4947,9 +4950,12 @@ components:
         config:
           type: object
           description: An object that defines configuration values for the trigger.
+        function_id:
+          type: string
+          description: The ID of the function that the trigger calls when it fires. This value is the same as and interchangeable with `event_processors.FUNCTION.function_id`.
```
mongodben, line 4955: what do you mean by "interchangeable with"?

nlarew: Basically you can either define these root-level fields or the event processor and the backend will duplicate the values to the one you didn't define. Kinda bad API design IMO but it is what it is.

Any thoughts on an alternative wording?

mongodben: oh wow, that's what i thought you meant, but figured it couldn't be right. 

given it's rather odd API design, i think it's worth explaining in more explicit detail. 

perhaps something like:

```md
This value is the same as `event_processors.FUNCTION.function_id`.
You can either define the value here or in `event_processors.FUNCTION.function_id`.
The App Services backend duplicates the value to the configuration location where you did not define it.

For example, if you define `function_id`, the backend duplicates it to `event_processors.FUNCTION.function_id`.
```

and then repeat this type of thing in all the places where it says 'interchangeable' now.

it's repetitive and clunky, but it's an API spec so doesn't need to be eloquent just functional and clear.

mongodben (approved):
lgtm!

PR #213, opened 2022-11-07
Remove iOS template note

cbush:
- The iOS UIKit template client is no longer compatible with the SwiftUI backend. The former uses PBS while the latter uses FS.
- TBD whether to support UIKit + FS.

### Jira

- 

- 

### Review Guidelines


dacharyc (approved):
LGTM

PR #220, opened 2022-11-09
Docsp-25435: Triggers during restore

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-25435

https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-25435/triggers/disable/#restoring-from-a-snapshot

### DIFF
--- a/source/triggers/disable.txt
+++ b/source/triggers/disable.txt
@@ -103,3 +103,26 @@ importing an application directory with {+cli-bin+}.
 
                realm-cli push
 
+
+Restoring from a Snapshot
+-------------------------
+
+Consider the following scenario:
+
+- A database trigger is disabled or suspended.
+
+- New documents are added while the trigger is disabled.
+
+- The database is restored from a snapshot to a time prior to the new documents 
+  being added.
+
+- The database trigger is restarted.
+
+In this case, the trigger will pick up all of the new file additions and fire a 
+notification for each. It will not resend prior notifications. 
+
+.. note:: Title
+
+   If a database trigger is enabled during snapshot restoration, you will see an 
+   error in the Edit Trigger section of the Atlas UI because it cannot connect to 
+   the Atlas cluster during the restore process. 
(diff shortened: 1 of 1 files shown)


source/triggers/disable.txt
```diff
@@ -103,3 +111,26 @@ importing an application directory with {+cli-bin+}.
...
+
+#. New documents are added while the trigger is disabled.
+
+#. The database is restored from a snapshot to a time prior to the new documents 
+   being added.
+
+#. The database trigger is restarted.
+
+In this case, the trigger will pick up all of the new file additions and fire a 
+notification for each. It will not resend prior notifications.
```
nlarew, line 130: I _think_ I get this but I'm not entirely sure. In my head this section kinda reads like "if you restore from a snapshot everything should just work" so I wanna make sure we're not missing something.

Would it be safe to say something like "the updates are applied on top of the snapshot and the trigger is able to pick up the change events for them"?

source/triggers/disable.txt
```diff
@@ -103,3 +111,26 @@ importing an application directory with {+cli-bin+}.
...
+#. The database trigger is restarted.
+
+In this case, the trigger will pick up all of the new file additions and fire a 
+notification for each. It will not resend prior notifications. 
+
+.. note::
+
+   If a database trigger is enabled during snapshot restoration, you will see an 
+   error in the Edit Trigger section of the Atlas UI because it cannot connect to 
+   the Atlas cluster during the restore process.
```
nlarew, line 136: Will this resolve itself or should we add guidance like "wait for the snapshot restoration to complete and then try to enable the trigger"

nlarew: Just checked the ticket, I think I misread this and we could clear it up. I read "is enabled" as "if you re-enable a disabled trigger".

Do you think we could make it clearer that an already enabled trigger will remain enabled but won't be able to connect until the snapshot restore is complete?

nlarew (approved):
LGTM + suggested extra sentence

PR #248, opened 2022-11-28
(DOCSP-26723): Add Recipes for Device Sync

dacharyc:
## Pull Request Info

I've fielded several questions from Swift developers over the last week or two about high-level process for getting from MongoDB Atlas data onto a client device, or getting data from a client device onto other client devices. I summarized these steps for them, but ended up linking out to a bunch of different documents to give them some overview of what to do. This seemed like a good case for recipes, so wanted to throw these together to see how we feel about them as a team. This information basically already existed, but wanted to collect it together into a single document that rolls everything up.

Not sure why the autobuilder check is showing as failed, but here is the build log with no issues: [Build Log](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=63990547dc742b91f23c2d46)

### Jira

- https://jira.mongodb.org/browse/DOCSP-26723

- [Sync Data in Atlas with a Client Application](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/add-recipes/sync/app-builder/sync-data-in-atlas-with-client/)
- [Sync Client Data with Atlas](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/add-recipes/sync/app-builder/sync-data-in-client-with-atlas/)
- [Stream Data Unidirectionally from a Client to Atlas](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/add-recipes/sync/app-builder/stream-data-from-client-to-atlas/)

### Review Guidelines


mongodben:
don't have time to do a full review on this (sorry), but a question: i see that the new pages are in a directory `source/recipes` (top lvl), but show up in the IA under `source/sync`.

why is this the structure? should the pages be moved under different directory `source/sync/recipes`?

dacharyc:
> question: i see that the new pages are in a directory `source/recipes` (top lvl), but show up in the IA under `source/sync`.
> 
> why is this the structure? should the pages be moved under different directory `source/sync/recipes`?

Yeah, good question, @mongodben . I originally added "Recipes" to the top-level nav for App Services, and had these pages there. After writing them up, I thought about how they're all Device Sync specific, so thought folks might expect to find them under Device Sync, and moved them to there in the ToC. But I'm also not super sold on having a "Recipes" nav element at all - I don't know if users will find that to be a meaningful title, and maybe they should be called something else. (Use Cases?) And I'm still undecided whether they should live under Device Sync, or be exposed at a higher level, alongside other "Recipes" or "Use Cases" - so that users don't necessarily need to know what feature set to use, but are just looking for something that tells how to do the task/use case they want to do.

So I guess TL;DR; I'm looking for opinions, and am happy to move the final structure to wherever we think it belongs, and call it whatever we think it should be called.

mongodben:
> So I guess TL;DR; I'm looking for opinions, and am happy to move the final structure to wherever we think it belongs, and call it whatever we think it should be called.

i don't have any opinion here.

though, since this is useful content and all related to sync stuff, putting in sync section seems reasonable, regardless of url structure.

cbush:
My initial thought: if we have a category in the TOC called "Recipes", we're moving back in the direction of basing the organization on content types rather than actions. I definitely like the content itself, so I will give more thought to where it could go.

source/sync/get-started/stream-data-from-client-to-atlas.txt
```diff
@@ -0,0 +1,267 @@
...
+      it, link it to an Atlas cluster, and specify the deployment model and
+      deployment region that work best for your application.
+
+      You can create an App Services App via the :ref:`Atlas UI 
+      <create-app-ui>`, the :ref:`Atlas App Services Command Line Interface 
+      <create-app-cli>`, or the :ref:`App Services Admin API <create-app-api>`.
+
+   .. step:: Configure Device Sync and Asymmetric Sync
+
+      Device Sync is the service that manages synchronizing data to Atlas
```
cbush, line 48: Maybe?
```suggestion
      Device Sync is the service that synchronizes data to Atlas 
```

source/sync/get-started/stream-data-from-client-to-atlas.txt
```diff
@@ -0,0 +1,267 @@
...
+      You can create an App Services App via the :ref:`Atlas UI 
+      <create-app-ui>`, the :ref:`Atlas App Services Command Line Interface 
+      <create-app-cli>`, or the :ref:`App Services Admin API <create-app-api>`.
+
+   .. step:: Configure Device Sync and Asymmetric Sync
+
+      Device Sync is the service that manages synchronizing data to Atlas 
+      and across your client devices. Device Sync, together with the 
+      Realm SDKs, automatically manages network connectivity, user 
+      authentication, and user permissions and data access.
```
cbush, line 51: User authentication -> not entirely true? The app services does that part.

Also not sure what you mean exactly by "data access".

source/sync/get-started/stream-data-from-client-to-atlas.txt
```diff
@@ -0,0 +1,267 @@
...
+
+      When you configure Device Sync, you specify the data source that the 
+      client devices can access, as well as the permissions that determine
```
cbush, line 54:
```suggestion
      client devices can access as well as the permissions that determine
```

source/sync/get-started/stream-data-from-client-to-atlas.txt
```diff
@@ -0,0 +1,267 @@
...
+
+      - Generate an App Services schema from your Realm object model. 
+        Refer to: :ref:`create-schema-from-rom`.
+      - Manually create an App Services Schema. For more information, refer 
+        to: :ref:`enforce-a-schema`.
+
+   .. step:: Enable an Authentication Provider
+
+      Your client devices must authenticate in order to access synced data.
+      App Services provides several authentication providers to enable
```
cbush, line 80: It might help to say "...such as email/password or anonymous..."

source/sync/get-started/sync-data-in-atlas-with-client.txt
```diff
@@ -0,0 +1,311 @@
...
+            :tabid: flutter
+
+            :ref:`Authenticate Users - Flutter SDK <flutter-authenticate>`
+
+         .. tab::
+            :tabid: kotlin
+
+            :ref:`Authenticate Users - Kotlin SDK <kotlin-authenticate>`
+
+   .. step:: Open a Synchronized Realm
```
cbush, line 182: It seems the catchphrase is _Synced realm_. I could be wrong?

cbush (approved):
First of all, sorry for the delay in getting to a serious review. I really love the content here. My only real hangup is where to put them. If it's in Getting Started, it creates an overwhelming feeling.

Here's what I suggest: a category between "Get Started" and "Configure & Update..." called... something like...

- "Fast Track"
- "App Builder's Resources"
- "Scenarios"

Move "Add Sync to a Local-Only App" and possibly "Flex Sync Permissions Guide" in there.

(Fully aware that my name suggestions might contradict the previous qualm with "Recipes" as being too much about the content type, but I'm thinking we can work around this by including the purpose/intended audience of the content (i.e. "fast", "app builder", "real world", etc etc)...)

cbush (approved):
LGTM. Let's merge and publish. We can see what the response is to the category title/placement and see if it needs adjustment later.

PR #255, opened 2022-12-02
(DOCSP-26793): Expand metrics

cbush:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26793

### Staged Changes

- https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/metrics-xpand/admin/api/v3/#tag/metrics
- https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/metrics-xpand/reference/metrics/
- 
### Reminder Checklist

### Review Guidelines


source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+   :class: singlecol
+
+App Services provides a variety of metrics related to App Services and Device
+Sync. You can access your metrics data with the Admin API's
+:admin-api-endpoint:`getMetrics <tag/metrics/operation/adminGetMetrics>` endpoint.
+
+The content on this page is provided for your information. The exact metrics
+returned may be subject to change.
+
+.. Source derived from https://docs.google.com/spreadsheets/d/1cjQopjPUsEKde6TlqO38B0Ko6n4vTwl4tgJ-LHSbf7Y/edit#gid=1354315462
```
mongodben, line 20: should this be deleted?

source/openapi-admin-v3.yaml
```diff
@@ -2246,8 +2239,10 @@ paths:
       operationId: adminGetMetrics
       summary: Retrieve App Services metrics
       description: |-
-        Retrieves a variety of metrics, including: Compute Time, Data
-        Transfer, Sync Minutes, Requests Succeeded, Requests Failed.
+        Retrieves a variety of metrics, including: Compute Time, Data Transfer,
+        Sync Minutes, Requests Succeeded, Requests Failed. For a complete list
+        of available metrics, see [App
+        Services Metrics Reference](https://mongodb.com/docs/atlas/app-services/reference/metrics)
```
mongodben, line 2245:
```suggestion
        Services Metrics Reference](https://mongodb.com/docs/atlas/app-services/reference/metrics).
```

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+    - Active open sync sessions.
+
+  * - AUTH_EGRESS_BYTES
+    - BYTES_PER_SECOND
+    - The amount of data transferred from the server due to
+      authentication requests.
+
+  * - AUTH_FAILED_REQUESTS
+    - SCALAR_PER_SECOND
+    -  failed authentication requests.
```
mongodben, line 65:
```suggestion
    -  Failed authentication requests.
```

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+App Services provides a variety of metrics related to App Services and Device
+Sync. You can access your metrics data with the Admin API's
+:admin-api-endpoint:`getMetrics <tag/metrics/operation/adminGetMetrics>` endpoint.
```
mongodben, line 15: nit: i don't think we refer to methods by their operationId elsewhere in the docs. 

i'd just put a plain-text description, smthn like
```rst 
:admin-api-endpoint:`Retrieve App Services metrics <tag/metrics/operation/adminGetMetrics>`
```

(also didn't know we had a `:admin-api-endpoint:` role)

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+
+The content on this page is provided for your information. The exact metrics
+returned may be subject to change.
```
mongodben, line 18: what's the purpose of the first sentence "The content on this page is provided for your information."?

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+A **scalar** is a singular value representing a magnitude or count. Its exact
+interpretation depends on the metric.
+
+A **scalar per second** represents a count over time ("rate"). The value is
+effectively the total count in the requested time window divided by the number
+of seconds in the window.
+
+Some scalar per second values are expressed as a 95th percentile. This means
+that the reading for the given metric is below the given value 95% of the time
+and above it the other 5% of the time.
```
mongodben, line 41: out of curiosity, do you know why this is the case?

seems weird to me that we're telling user that our metric is this good 5% or more of the time

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+   :header-rows: 1
+   :widths: 50 20 30
+
+  * - Metric Name
+    - Unit
+    - Description
+
+  * - ACTIVE_OPEN_SYNC_SESSIONS
+    - SCALAR
+    - Active open sync sessions.
```
mongodben, line 56: can you be more descriptive? 

is this the total number of subscriptions that were open over a given duration? 
including ones that were already open? excluding?

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+    - SCALAR_PER_SECOND
+    - Rate of successful logins.
+
+  * - AUTH_SUCCESSFUL_REQUESTS
+    - SCALAR_PER_SECOND
+    - Rate of successful authentication requests.
+
+  * - AUTH_TOTAL_USERS
+    - SCALAR
+    - The total number of users.
```
mongodben, line 81: how does this work with the time span in the query params? 

i/e number of users at beginning of time span could be different from the number at the end

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+    - Rate of failed function call requests from client SDKs.
+
+  * - SDK_FNS_RESPONSE_MS
+    - MILLISECONDS
+    - 95th percentile response time for client SDK function calls.
+
+  * - SDK_MQL_COMPUTE_MS
+    - SCALAR_PER_SECOND
+    - The number of milliseconds per second spent computing client SDK remote
+      MongoDB access requests.
```
mongodben, line 167: ditto https://github.com/mongodb/docs-app-services/pull/255/files#r1041320654

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+  * - SYNC_CLIENT_UPLOADS_INVALID
+    - SCALAR_PER_SECOND
+    - Device Sync client invalid uploads.
+
+  * - SYNC_CURRENT_OPLOG_LAG_MS_SUM
+    - MILLISECONDS
+    - The sum of Device Sync oplog lag in milliseconds.
+
+  * - SYNC_EGRESS_BYTES
+    - BYTES_PER_SECOND
```
mongodben, line 203: should this just be bytes?

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+  * - TRIGGERS_COMPUTE_MS
+    - SCALAR_PER_SECOND
+    - Rate of compute time spent for Triggers.
+
+  * - TRIGGERS_CURRENT_OPLOG_LAG_MS_SUM
+    - MILLISECONDS
+    - Sum of Triggers oplog lag.
+
+  * - TRIGGERS_EGRESS_BYTES
+    - BYTES_PER_SECOND
```
mongodben, line 256: again should just be bytes, not bytes/second?

mongodben (commented):
back to you!

source/reference/metrics.txt
cbush: Why?

source/reference/metrics.txt
cbush: Thanks, this is a search and replace flub. It's actually supposed to say Rate of ...

source/reference/metrics.txt
cbush: Just a preamble to the next thing. But even that's questionable - eng says they might be private/subject to change but product would prefer they don't change.

source/reference/metrics.txt
cbush: It's certainly inconsistent. Usually we refer to the category. I'm referring to a specific endpoint here, so I don't see the benefit of calling it something other than what it is. But I don't feel strongly.

source/reference/metrics.txt
cbush: "Good" implies the metric is a value judgment and that higher is always better. It's neither. The 95th percentile scalar/sec type is used for things like response times where lower is actually better. It's just saying that MOST of the time, it's below X, but sometimes it's above X. I don't know the exact technical reason why it's presented this way.

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+    - SCALAR_PER_SECOND
+    - Rate of successful authentication requests.
+
+  * - AUTH_TOTAL_USERS
+    - SCALAR
+    - The total number of users.
+
+  * - ENDPOINTS_COMPUTE_MS
+    - SCALAR_PER_SECOND
+    - The number of milliseconds per second spent computing endpoint requests.
```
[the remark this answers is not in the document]
cbush: No. Note that the type is SCALAR_PER_SECOND, not MILLISECOND (or scalar). See the note about scalar per second. Basically this says how much time out of every second in the time window was spent computing. Really tried to keep this clean and clear but let me know if you have suggestions for how to clarify more, as I'm sure you're not the only one who would be left scratching your head.

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+    - SCALAR_PER_SECOND
+    - Overall rate of failed requests.
+
+  * - OVERALL_SUCCESSFUL_REQUESTS
+    - SCALAR_PER_SECOND
+    - Overall rate of successful requests.
+
+  * - OVERALL_SYNC_MINUTES
+    - SCALAR_PER_SECOND
+    - Overall sync minutes.
```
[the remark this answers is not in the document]
cbush: This is not supposed to be Total. This is a scalar per second value, which means it's actually way more confusing. Unlike the "milliseconds per second" construct above, this is minutes per second, which makes no sense...

source/reference/metrics.txt
```diff
@@ -0,0 +1,270 @@
...
+
+  * - SDK_COMPUTE_MS
+    - SCALAR_PER_SECOND
+    - The number of milliseconds per second spent computing client SDK functions calls.
+
+  * - SDK_EGRESS_BYTES
+    - BYTES_PER_SECOND
+    - The amount of data transferred from the server due to SDK functions calls.
```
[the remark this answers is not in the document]
cbush: No. This comes from the source and from empirical testing. Almost everything has to do with rates as opposed to total values. But I can see I might need to clarify the phrases on the BYTES_PER_SECOND ones.

source/reference/metrics.txt
mongodben: i was told at some point to delete comments, but i have no personal issue w keeping. seems vaguely useful for future maintainers of the docs, assuming that spreadsheet is kept up to date.

source/reference/metrics.txt
mongodben: but the operationId isn't exposed in the spec (besides the URL path), so don't think it should be used here. recommending: 

```suggestion
:admin-api-endpoint:`Retrieve App Services metrics <tag/metrics/operation/adminGetMetrics>` endpoint.
```

source/reference/metrics.txt
mongodben: thanks for indulging my curiosity

source/reference/metrics.txt
```diff
@@ -0,0 +1,269 @@
...
+   :header-rows: 1
+   :widths: 50 20 30
+
+  * - Metric Name
+    - Unit
+    - Description
+
+  * - ACTIVE_OPEN_SYNC_SESSIONS
+    - SCALAR
+    - The number of active, open sync sessions at the time of the reading.
```
mongodben, line 55: what do you mean by "time of reading"? is this not using the query params for `start` and `end`?

source/reference/metrics.txt
```diff
@@ -0,0 +1,269 @@
...
+    - SCALAR_PER_SECOND
+    - The rate of successful logins.
+
+  * - AUTH_SUCCESSFUL_REQUESTS
+    - SCALAR_PER_SECOND
+    - The rate of successful authentication requests.
+
+  * - AUTH_TOTAL_USERS
+    - SCALAR
+    - The number of existing users at the time of the reading.
```
mongodben, line 80: ditto https://github.com/mongodb/docs-app-services/pull/255/files#r1042559324

source/reference/metrics.txt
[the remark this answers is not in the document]
mongodben: thanks for clarifying. i think the explanation above is good. it's a funky concept, and while it certainly could be expanded upon more, i don't know how necessary that would be for this fairly small (and presumably not highly trafficed) documentation page

mongodben (commented):
back to you! looking good, just a couple of small clarification points before approval

source/reference/metrics.txt
cbush: Oh, I see what you're saying. Got it.

source/reference/metrics.txt
cbush: That was the idea. I usually say to delete things that are "commented out", not "comments".

source/reference/metrics.txt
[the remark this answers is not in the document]
cbush: We'll see if we get feedback on it. I could go with what I did for the really weird one "minutes per second". I said it's the number of minutes in the window divided by the number of seconds in that window. However, the problem(?) with that is that I'm then talking about "windows".

source/reference/metrics.txt
mongodben: > If the interval is like a day, is it the number of sessions open that day or is it a snapshot of the exact time of measurement? Not quite sure about this.

i think this is an important point to clarify.

> I think throwing in the word "reading" was a mistake. I can call it "measurement". Not sure if that alone would clarify anything but at least I wouldn't be using a new term. (FWIW, I chose the former because it sounded less stuffy but I changed my mind looking at it now.)

i'm partial to measurements..but maybe i'm just stuffy. defer to you on phrasing there. i think the above clarification point is what must be clarified for this PR.

mongodben (approved):
i think the addition clarifies it well. nice work here.

lgtm!

PR #266, opened 2022-12-15
(DOCSP-26974): Add end-to-end example for a custom HTTPS endpoint function

davidhou17:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26974

### Staged Changes

- [Custom HTTPS Endpoints](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-26974/data-api/custom-endpoints/#write-an-endpoint-function)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/data-api/custom-endpoints.txt
+++ b/source/data-api/custom-endpoints.txt
@@ -136,16 +136,18 @@ deploying configuration files with Realm CLI:
          JSON or EJSON. You can enable :guilabel:`Respond With Result`
          to automatically include the endpoint function's return value
          as the response body.
+
+      5. Optionally, specify or write an :ref:`endpoint function <endpoint-function>`.
       
-      5. For additional security, you can configure :ref:`request
+      6. For additional security, you can configure :ref:`request
          authorization <endpoint-authorization>`.
       
-      6. Save the Data API configuration.
+      7. Save the Data API configuration.
       
-      6. Configure :ref:`access permissions <data-api-permissions>` to
+      8. Configure :ref:`access permissions <data-api-permissions>` to
          allow requests to securely read and write data.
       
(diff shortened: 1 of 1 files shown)


source/data-api/custom-endpoints.txt
```diff
@@ -343,55 +345,26 @@ Write an Endpoint Function
...
 function, you can import libraries from npm, connect to a linked MongoDB
-Atlas cluster, and call other serverless functions.
+Atlas cluster, and call other serverless functions. 
 
-.. tip::
-   
-   You define and edit endpoint handler functions on the
-   :guilabel:`Functions` page of the App Services UI or in the :ref:`/functions
-   <appconfig-functions>` folder of an app configuration directory.
+To define and edit endpoint handler functions:
```
nlarew, line 350: Let's be clearer that these are all alternative ways to do the same thing. Right now it reads sort of like a procedure despite using an unordered list.

- You can define and edit endpoint handler functions from the UI, CLI, _or Admin API_
- Depending on your workflow, you can define functions... (bulleted list below this)

davidhou17: Good point! I thought of a workaround in the most recent version - I emphasized that you can create a function in the same page while creating a custom endpoint, and I listed the alternative methods below. 

Do you think it feels excessive to describe each alternative? I'm on the fence. The alternative would be something like :

`To define a new function when creating an endpoint
 in the App Services UI, select the + New Function option in the Function section. Depending on your workflow, you can also define and edit endpoint handler functions from the UI, CLI, or Admin API. For more information, see Define a Function or App Services Admin API.`

nlarew (approved):
LGTM

PR #289, opened 2023-01-07
Clean up files (Test)

cbush:
https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/cleanup


source/migrate.txt
```diff
@@ -1,60 +0,0 @@
-=====================================
-Migrate Your App from Stitch to Realm
```
dacharyc, line 2: I wonder if this page and the one below it - stitch.txt - should continue to exist? I don't know how long we plan/need to support migrating from Stitch to Realm.

source/sync/event-library.txt
```diff
@@ -1,465 +0,0 @@
-.. _app-services-event-library:
-
-=============
-Event Library
```
dacharyc, line 4: This should definitely _not_ be deleted. There used to be a link to it from the Device Sync ToC - it looks like we lost this from the ToC during the Sync section refactor: https://github.com/mongodb/docs-app-services/pull/191

Maybe a good place for this to live would be "App Builders Resources"?

dacharyc (changes requested):
Overall this looks like a nice cleanup pass, but there's one file that definitely shouldn't be deleted, and a couple that I have a question about.

source/migrate.txt
cbush: This one exists on the Realm side so I figured we could redirect.

source/sync/event-library.txt
cbush: At least this proves the script will reveal mistakes like that!

dacharyc (approved):
LGTM!

PR #315, opened 2023-01-20
Tutorial v2 updates

cbush:
### Staged Changes

- https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/tutorial-v2-updates/


dacharyc (approved):
LGTM, although I spotted that the Maui tutorial is missing the updates to the "Start with the Template" section that the other tutorials got. We should add that before adding it back on Monday.

PR #332, opened 2023-01-30
(DOCSP-26937) [Port] Control Developer Access (/security/developer-access)

nlarew:
### Jira

- (DOCSP-26937) [Port] Control Developer Access (/security/developer-access)

### Staged Changes

- [Secure Your App > Control Developer Access](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/ia-security/security/developer-access/)

### DIFF
--- a/source/security.txt
+++ b/source/security.txt
@@ -16,12 +16,12 @@ Secure Your App
 .. toctree::
    :titlesonly:
    :hidden:
- 
+
    Configure Network Security </security/network>
+   Control Developer Access </security/developer-access>
 
 .. TO ADD LATER:
    Authorize Incoming Requests </security/authorization>
-   Control Developer Access </security/developer-access>
 
 Overview
 --------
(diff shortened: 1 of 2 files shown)


source/security.txt
```diff
@@ -16,12 +16,12 @@ Secure Your App
 .. toctree::
    :titlesonly:
    :hidden:
- 
+
    Configure Network Security </security/network>
+   Control Developer Access </security/developer-access>
```
dacharyc, line 21: This is just a nit, and I'm not sure whether you covered this in your tree test, but I wonder if this ToC item/page name should be "Configure Developer Permissions"? I think that better parallels "Configure Network Security", and "control" just rubs me the wrong way for some reason. If we do want to emphasize the restriction aspect of this, maybe we could consider a name like "Limit Developer Access" or "Restrict Developer Access"?

dacharyc (approved):
This looks good - just a Q about naming.

PR #334, opened 2023-02-01
DOCSP-27674 Add the after query parameter for pending users API

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-27674

### Staged Changes

- [Atlas App Services API](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-27674/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -1701,6 +1701,18 @@ paths:
     parameters:
       - "$ref": "#/components/parameters/GroupId"
       - "$ref": "#/components/parameters/AppId"
+      - in: query
+        name: after
+        required: false
+        schema: 
+          type: string
+        description: |-
+          The unique ``_id`` for a pending user. The pending users API
+          returns a max of 50 documents, or pending users, at a time. The 
+          ``after`` parameter can be used to view additional pending users. 
+          For example, by providing the ``_id`` of the 50th pending user for 
+          the ``after`` parameter, you can view the next 50 pending users. 
+
     get:
       tags:
         - users


MongoCaleb (commented):
I wonder if this deserves a bit broader editing? I think, for example, that the definition of "List pending users" should say "List pending user account registrations. Up to 50 pending users are returned in a call." And then the definition of the "after" query param simplifies to something like "If there are more than 50 pending users, pass the _id of the last user in the previous response to the "after" query parameter in this request." or some such nonsense. I think it'll make it all clearer.

source/openapi-admin-v3.yaml
```diff
@@ -1701,12 +1701,23 @@ paths:
     parameters:
       - "$ref": "#/components/parameters/GroupId"
       - "$ref": "#/components/parameters/AppId"
+      - in: query
+        name: after
+        required: false
+        schema: 
+          type: string
+        description: |-
+          The unique ``_id`` for a pending user. List pending users can return 50
```
MongoCaleb, line 1710: A suggestion (only) that I think makes it a bit more scan-able:

To view additional results, use the ``_id`` of the last pending user listed 
 in the response to call ``list pending users`` again, passing the _id it to the ``after`` parameter.

MongoCaleb (approved):
I'm going to LGTM this since I'll be out tomorrow. Please consider -- but don't feel like you need to implement -- my one remaining suggestion.

MongoCaleb:
Love the changes!

PR #336, opened 2023-02-02
(DOCSP-27741) Bearer Auth Example for Custom & Generated Endpoints

nlarew:
### Jira

- (DOCSP-27741) Bearer Auth Example for Custom & Generated Endpoints

### Staged Changes

- [Generated Endpoints > Authenticate the Request](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/api-auth/data-api/generated-endpoints/#authenticate-the-request)
- [Custom Endpoints > Authenticate the Request](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/api-auth/data-api/custom-endpoints/#authenticate-the-request)

PR #337, opened 2023-02-03
(DOCSP-27664) Update top Data API landing to reflect new positioning / frontend access

nlarew:
## Pull Request Info

- Updates Data API landing page to focus more on browser use case + expand conceptual/technical explanations.
- Moves data formats from the landing page to a separate sub-page
- Note: this is rebased on top of https://github.com/mongodb/docs-app-services/pull/336

### Jira

- (DOCSP-27664) Update top Data API landing to reflect new positioning / frontend access

### Staged Changes

- [Data API](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/api-auth-2/data-api/)
- (Moved) [Data API > Data Formats](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/api-auth-2/data-api/data-formats/)

### DIFF
--- a/source/data-api.txt
+++ b/source/data-api.txt
@@ -27,15 +27,15 @@ You can use the Data API to connect to MongoDB Atlas from any platform
 that supports HTTPS, including:
 
 - Web browsers
-- Servers
+- Web servers
 - CI/CD pipelines
 - Serverless & Edge compute environments
 - Mobile applications
 - Internet-Of-Things devices
 
 You don't need to install any database drivers or opinionated libraries
-to work with the Data API. Instead, you send standard HTTPS requests like the
-following:
+to work with the Data API. Instead, you send standard HTTPS requests
+like the following:
 
 .. io-code-block::
    
@@ -73,11 +73,11 @@ The Data API supports two types of endpoints:
   To learn more, including how to set up and call endpoints, see
   :ref:`Data API Endpoints <data-api-endpoints>`.
 
-- **Custom Endpoints** are app-specific API routes handled by Atlas Functions
-  that you write. You can use custom endpoints to define operations that
-  fit your use case specifically. For example, you could create an
-  endpoint that runs a pre-defined aggregation or that integrates with
-  an external webhook service.
+- **Custom Endpoints** are app-specific API routes handled by Atlas
+  Functions that you write. You can use custom endpoints to define
+  operations that fit your use case specifically. For example, you could
+  create an endpoint that runs a pre-defined aggregation or that
+  integrates with an external webhook service.
 
   To learn more, including how to create and call endpoints, see
   :ref:`Custom HTTPS Endpoints <custom-endpoints>`.
@@ -121,19 +121,13 @@ For each incoming request, the Data API:
 When to Use the Data API
 ------------------------
 
-An operation called through a Data API endpoint will likely take longer
-than the corresponding MongoDB operation called through a connected
-MongoDB Driver. For high-load use-cases and latency sensitive
-applications, we recommend connecting directly to your database with a
-MongoDB driver. To learn more, visit the :driver:`MongoDB Drivers </>`
-documentation.
-
-
-TODO
-
-
-"Using the MongoDB Drivers to build on top of Atlas is still the general recommended approach (especially when connecting from a Server) as it will be more performant and flexible."
-
+For server applications, and especially for high-load and latency
+sensitive use-cases, we recommend connecting directly to Atlas with a
+MongoDB driver. Operations called through a Data API endpoint take
+longer to complete than the corresponding MongoDB operations called
+through a driver. Additionally, the drivers provide more flexibility and
+control over how your operations are executed. To learn more, visit the
+:driver:`MongoDB Drivers </>` documentation.
 
 We recommend using the Data API when:
 
@@ -144,8 +138,9 @@ We recommend using the Data API when:
   environment. For example, some edge compute environments don't support
   database drivers or connection pooling.
 
-- You want to develop a new feature and prefer to work on the client
-  side first before later creating and refining the API layer.
+- You want to develop a new feature and prefer a flexible solution for
+  working on the client side first before later creating and refining
+  the API layer.
 
 - You want to integrate Atlas data access into a federated API gateway.
 
--- a/source/data-api.txt
+++ b/source/data-api.txt
@@ -17,10 +17,14 @@ Atlas Data API
 .. contents:: On this page
    :depth: 2
 
-The Data API is a managed service that lets you securely work with
-data stored in Atlas using standard HTTPS requests. You can use the Data
-API to connect to MongoDB Atlas from any platform that supports HTTPS,
-including:
+The Data API is a managed service that lets you securely work with data
+stored in Atlas using standard HTTPS requests. The Data API is not a
+direct connection to your database. Instead, the API is a fully-managed
+middleware service that sits between your cluster and the clients that
+send requests.
+
+You can use the Data API to connect to MongoDB Atlas from any platform
+that supports HTTPS, including:
 
 - Web browsers
 - Servers
@@ -28,10 +32,6 @@ including:
 - Serverless & Edge compute environments
 - Mobile applications
 - Internet-Of-Things devices
-
-The Data API is not a direct connection to your database. Instead, the
-API is a fully-managed middleware service that sits between your cluster
-and the clients that send requests.
 
 You don't need to install any database drivers or opinionated libraries
 to work with the Data API. Instead, you send standard HTTPS requests like the
@@ -121,6 +121,20 @@ For each incoming request, the Data API:
 When to Use the Data API
 ------------------------
 
+An operation called through a Data API endpoint will likely take longer
+than the corresponding MongoDB operation called through a connected
+MongoDB Driver. For high-load use-cases and latency sensitive
+applications, we recommend connecting directly to your database with a
+MongoDB driver. To learn more, visit the :driver:`MongoDB Drivers </>`
+documentation.
+
+
+TODO
+
+
+"Using the MongoDB Drivers to build on top of Atlas is still the general recommended approach (especially when connecting from a Server) as it will be more performant and flexible."
+
+
 We recommend using the Data API when:
 
 - You want to run MongoDB operations from a web application or other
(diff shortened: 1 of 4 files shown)


source/data-api.txt
```diff
@@ -12,32 +12,53 @@ Atlas Data API
...
+- Serverless & Edge compute environments
+- Mobile applications
+- Internet-Of-Things devices
+
+The Data API is not a direct connection to your database. Instead, the
+API is a fully-managed middleware service that sits between your cluster
+and the clients that send requests.
+
+You don't need to install any database drivers or opinionated libraries
+to work with the Data API. Instead, you send standard requests like the
```
mongodben, line 37:
```suggestion
to work with the Data API. Instead, you send standard HTTPS requests like the
```

source/data-api.txt
```diff
@@ -12,32 +12,53 @@ Atlas Data API
...
-and client apps. For each incoming request, the Data API authenticates
-the calling user, authorizes the request, runs the endpoint logic, and
-then returns an HTTPS response to the caller.
+   .. output::
+      :language: json
+
+      { "insertedId": "63dc56ac74ddb86ed3eb8474" }
 
 Endpoints
 ---------
```
mongodben, line 64: i don't think it's smthn to address in this PR, but i find it kinda funny that we're calling both generated endpoints and custom endpoints 'the data api', but then also the generated endpoints are called 'the data api'.

probably this is smthn that extends to product/product marketing. but worth bringing up imo.

curious your thoughts.

source/data-api.txt
```diff
@@ -53,297 +74,71 @@ The Data API supports two types of endpoints:
   :ref:`Data API Endpoints <data-api-endpoints>`.
 
 - **Custom Endpoints** are app-specific API routes handled by functions
```
mongodben, line 76:
```suggestion
- **Custom Endpoints** are app-specific API routes handled by Atlas Functions
```

source/data-api.txt
```diff
@@ -53,297 +74,71 @@ The Data API supports two types of endpoints:
...
-  </reference/mongodb-extended-json/>`, is a superset of standard JSON
-  that uses structured fields to represent BSON data that don't have
-  corresponding JSON types. This fully represents your data but requires
-  your client to understand how to work with EJSON.
+Data API requests may resemble traditional database operations, like
+``find`` or ``insertOne``, but the Data API is not a direct connection
+to your database. Instead, the Data API adds additional authentication,
+authorization, and correctness checks to ensure that your data is only
+accessed or modified in the ways you allow. This allows you to safely
+access data in Atlas from unsafe clients like web apps.
```
mongodben, line 93: to call web apps 'unsafe' rings a little funky to me.

i wonder if there's a better word that we could use here. perhaps smthn like: 

```suggestion
access data in Atlas from potentially vulnerable clients like web apps.
```

(got that suggestion w help of chat gpt 😉)

source/data-api.txt
```diff
@@ -53,297 +74,71 @@ The Data API supports two types of endpoints:
...
-          :copyable: false
+3. **Runs the requested operation.** This might involve reading or
+   writing data in Atlas with a generated endpoint or invoke a custom
+   function that you wrote.
 
-          "2022-01-12T02:33:23.067Z"
+   For requests that read or write data in Atlas, the Data API also
+   enforces the access control rules and document schemas defined in
+   your App. This means that users can only access data they're allowed
+   to read and that requests will fail if they include an invalid write
```
mongodben, line 112:
```suggestion
   your App. This means that users can only access data they're allowed
   to read and write. Requests fail if they include an invalid write
```

source/data-api.txt
nlarew: Yeah it's silly IMO but that's what we're calling them for now. I have some ideas on how we could be clearer that I've brought up with Sumedha but I'll reintroduce them with PM again.

source/data-api.txt
```diff
@@ -53,297 +74,71 @@ The Data API supports two types of endpoints:
...
 
-   * - ``Decimal128``
-     
-     - .. code-block:: json
-          :copyable: false
+4. **Returns an HTTPS response to the caller.** The response includes
+   the result of a generated endpoint operation or any data that you
+   return from a custom endpoint. Requests can choose to receive the
+   response in either :ref:`JSON or EJSON format
+   <data-api-data-formats>`.
```
mongodben, line 119: the last sentence reads a little funny to me b/c it seems to assign agency to the request. consider rephrase to assign agency to developer: 

```suggestion
   return from a custom endpoint. In the request, you can choose to receive the
   response in either :ref:`JSON or EJSON format
   <data-api-data-formats>`.
```

source/data-api.txt
```diff
@@ -52,298 +73,72 @@ The Data API supports two types of endpoints:
...
-          { <content> }
-     
-     - .. code-block::
-          :copyable: false
+- You can't or don't want to manage a MongoDB driver in your server-side
+  environment. For example, some edge compute environments don't support
+  database drivers or connection pooling.
 
-          { <content> }
+- You want to integrate Atlas data access into a federated API gateway
```
mongodben, line 133:
```suggestion
- You want to integrate Atlas data access into a federated API gateway.
```

source/data-api.txt
```diff
@@ -52,298 +73,72 @@ The Data API supports two types of endpoints:
...
 
-   * - ``Double``
-     
-     - .. code-block:: json
-          :copyable: false
+- You want to connect to App Services from an environment not currently
+  supported by a Realm SDK and don't want to use a driver to connect
```
mongodben, line 136: ref to realm sdks landing page here? https://www.mongodb.com/docs/realm/

tho on reading the landing page, it really positions realm as a mobile DB, rather than a mobile DB + client to access App Services, so reader might be confused if they're not familiar w realm and wind up there, so 🤷 

curious your thoughts here. 

obvi out of scope for this PR, but maybe we should have a page about realm on the App Services docs explaining what Realm SDKs are, and how they related to App Services.

source/data-api.txt
```diff
@@ -12,32 +12,53 @@ Atlas Data API
    Custom HTTPS Endpoints </data-api/custom-endpoints>
```
mongodben, line 12: this page is an excellent product overview!

source/data-api/data-formats.txt
```diff
@@ -0,0 +1,369 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+MongoDB stores data in a format called BSON, which is similar to a JSON
+object in structure but supports additional data types and uses a binary
+encoding. BSON is efficient for computers but is not human readable so
+you can't work with it directly.
```
mongodben, line 16:
```suggestion
encoding. BSON is efficient for computers but is not human readable,
so you can't work with it directly.
```

source/data-api/data-formats.txt
```diff
@@ -0,0 +1,369 @@
...
+Data Formats
+============
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+MongoDB stores data in a format called BSON, which is similar to a JSON
```
mongodben, line 13: maybe link word BSON to docs? 

```suggestion
MongoDB stores data in a format called `BSON <https://bsonspec.org/>`__,
which is similar to a JSON
```

source/data-api/data-formats.txt
```diff
@@ -0,0 +1,369 @@
...
+  understand. However, JSON cannot represent all BSON types so JSON
+  responses may lose type information for some fields. For example, BSON
+  has distinct types for 32-bit integers and 64-bit floats but a JSON
+  response represents both as a ``number``.
+
+- The **EJSON** format, short for :manual:`MongoDB Extended JSON
+  </reference/mongodb-extended-json/>`, is a superset of standard JSON
+  that uses structured fields to represent BSON data that don't have
+  corresponding JSON types. This fully represents your data but requires
+  your client to understand how to work with EJSON.
```
mongodben, line 31: doesn't need to be in this PR, but would be helpful to point users to docs on how to work w EJSON from client code.

mongodben (approved):
nice work here! like the new landing page a lot. clearly explains the value prop. 

couple small things and few questions throughout but nothing blocking.

source/data-api.txt
```diff
@@ -52,298 +73,72 @@ The Data API supports two types of endpoints:
...
-          {
-             "T": 1565545664,
-             "I": 1
-          }
+An operation called through a Data API endpoint will likely take longer
+than the corresponding MongoDB operation called through a connected
+MongoDB Driver. For high-load use-cases and latency sensitive
+applications, we recommend connecting directly to your database with a
+MongoDB driver. To learn more, visit the :driver:`MongoDB Drivers
+</data-api-data-formats>` documentation.
```
mongodben, line 144: just noticed this link is broken. fix before merge.

source/data-api.txt
```diff
@@ -12,32 +12,53 @@ Atlas Data API
...
 
-.. code-block:: bash
+- Web browsers
+- Servers
+- CI/CD pipelines
+- Serverless & Edge compute environments
+- Mobile applications
+- Internet-Of-Things devices
+
+The Data API is not a direct connection to your database. Instead, the
```
drewdipalma, line 32: Nit: I think this might help set the stage better either combined with the first sentence or as a second sentence.  Feels like we're splitting the definition in two here.

source/data-api.txt
```diff
@@ -52,298 +73,72 @@ The Data API supports two types of endpoints:
...
-          }
-     
-     - .. code-block:: json
-          :copyable: false
-
-          {
-             "T": 1565545664,
-             "I": 1
-          }
+An operation called through a Data API endpoint will likely take longer
```
drewdipalma, line 139: I would like us to be more specific here and add above the bulleted list.  Something along the lines of "Using the MongoDB Drivers to build on top of Atlas is still the general recommended approach (especially when connecting from a Server) as it will be more performant and flexible."

source/data-api.txt
drewdipalma, line 144: Another bullet that we might consider adding to this list – 

- "You are looking to create a new feature and would prefer starting your development on the client side and then moving to create / refine the API layer when your requirements are refined."

drewdipalma (approved):
Approving with some minor comments

source/data-api/data-formats.txt
nlarew: Don't think we actually have a great article on this. Not sure if we want it in this section or elsewhere but would be great to have.

source/data-api.txt
nlarew: > on reading the landing page, it really positions realm as a mobile DB, rather than a mobile DB + client to access App Services, so reader might be confused if they're not familiar w realm and wind up there

Link could be helpful. I think the paragraph that linked them there has enough context that some/most SDKs have an App Services integration.

> obvi out of scope for this PR, but maybe we should have a page about realm on the App Services docs explaining what Realm SDKs are, and how they related to App Services

Agreed - I thought we at least had this on the introduction page but we don't even really have it there, just a couple mentions of the SDKs in broader contexts.

PR #350, opened 2023-02-13
(DOCSP-26938) [Section] Define a Data Model

nlarew:
### Jira

- (DOCSP-26938) [Section] Define a Data Model

### Staged Changes

- [Define a Data Model](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/ia-data-model/data-model/)

### DIFF
--- a/source/data-model.txt
+++ b/source/data-model.txt
@@ -17,10 +17,10 @@ Overview
 --------
 
 Your App's data model is a description of the data that your App uses.
-Your data model helps to ensure that your data is well-formed, contains
+Your data model ensures that your data complies with a specific structure, contains
 only valid values, and only changes in ways that you expect.
 
-A data model is composed of several components, including:
+A data model consists of several components, including:
 
 - A set of declarative schemas that describe your App's data types.
 
@@ -53,7 +53,7 @@ Document Schemas
 ----------------
 
 A document schema is a JSON object that defines the structure of a
-particular type of data that your App uses. Document schemas control
+particular type of data that your App uses. Document schemas describe
 which fields a document has, the type of values those fields contain,
 and conditions that must be met for a change of value to be valid.
 
@@ -73,7 +73,7 @@ Relationships
 -------------
 
 A relationship is a connection between two document schemas that lets
-you logically associate one document with one or more other documents.
+you logically associate one document with zero or more other documents.
 
 In a relationship, a "source" document stores a reference to related
 documents, for example a list of related document ``_id`` values. When
--- a/source/data-model.txt
+++ b/source/data-model.txt
@@ -1,0 +1,118 @@
+.. _data-model:
+
+===================
+Define a Data Model
+===================
+
+.. contents:: On this page
+   :depth: 1
+
+.. toctree::
+
+   Schemas </schemas>
+   Relationships </schemas/relationships>
+   Values & Secrets </values-and-secrets>
+
+Overview
+--------
+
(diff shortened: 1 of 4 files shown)


source/data-model.txt
```diff
@@ -0,0 +1,118 @@
...
+
+   Schemas </schemas>
+   Relationships </schemas/relationships>
+   Values & Secrets </values-and-secrets>
+
+Overview
+--------
+
+Your App's data model is a description of the data that your App uses.
+Your data model helps to ensure that your data is well-formed, contains
```
dacharyc, line 20: Nit: the data model doesn't actually ensure that the data is "well-formed." If the data model isn't well-formed, the data won't be well-formed. It's probably more accurate to say that something about how it enforces a specific data structure.

```suggestion
Your data model ensures that your data complies with a specific structure, contains
```

source/data-model.txt
```diff
@@ -0,0 +1,118 @@
...
+only valid values, and only changes in ways that you expect.
+
+A data model is composed of several components, including:
```
dacharyc, line 23: Nit: passive voice.
```suggestion
A data model consists of several components, including:
```

source/data-model.txt
```diff
@@ -0,0 +1,118 @@
...
+
+- Automatically generate syncable Realm object models in your native
+  programming language using an App administration tool. To learn more,
+  see :ref:`Generate SDK Object Models <generate-sdk-object-models>`.
+
+Document Schemas
+----------------
+
+A document schema is a JSON object that defines the structure of a
+particular type of data that your App uses. Document schemas control
```
dacharyc, line 56: This is just a nit, but pedantically, I want something else here because a schema doesn't directly do the controlling/enforcing. It's just a blueprint, which our very complicated rules engine uses to enforce these things. 

```suggestion
particular type of data that your App uses. Document schemas describe
```

source/data-model.txt
```diff
@@ -0,0 +1,118 @@
...
+MongoDB. This includes inserts, updates, and deletes from an API
+service, Function, or Device Sync.
+
+To learn more, see :ref:`Schemas <schemas>`.
+
+Relationships
+-------------
+
+A relationship is a connection between two document schemas that lets
+you logically associate one document with one or more other documents.
```
dacharyc, line 76: Couldn't relationships technically have zero other documents? i.e. an empty list or optional to-one relationship?

I wonder if we should instead say "document type"/ "document types" in which case this would make sense?

```suggestion
you logically associate one document with zero or more other documents.
```

source/data-model.txt
```diff
@@ -0,0 +1,118 @@
...
+Values & Secrets
+----------------
+
+A value is a named constant that you can access by name in various ways
+throughout your App. For example, you can use a value to define the base
+URL of an external API service and then reference the value instead of
+duplicating the base URL across many Function definitions.
+
+You can also define an environment value, which can change depending on
+your App's environment tag. For example, you might use a different API
```
dacharyc, line 95: Should this link to the "environment" documentation? (And should above link to the "Function" documentation? I don't know where this lands on the spectrum of "unnecessary wiki-style link" vs "easily give users a link to related docs so they can understand the context."

dacharyc (approved):
A few small Qs and suggestions, but overall LGTM 👍

source/data-model.txt
nlarew: I don't think it's necessarily too much linking, will add

PR #357, opened 2023-02-15
Custom User Data updates

nlarew:
- Adds info about user creation function to the UI custom user data procedure
- Buffs up the organization of the page
- Adds Admin API procedures across the page

### Jira

- (DOCSP-27025) [BAAS] Add user creation function option in Custom User Data UI

### Staged Changes

- [Define Custom User Metadata](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/user-creation-fn/users/custom-metadata/)


source/authentication.txt
```diff
@@ -118,8 +118,9 @@ User Metadata
 Each authentication provider can associate metadata fields with an 
 application user. Some providers, such as Email/Password, always add 
 specific fields. Others allow you to configure the data to associate with 
-each user. Authentication providers include these metadata fields in a 
-user object:
+each user.
+
+To learn more, see :ref:`Authentication Provider Metadata <auth-provider-metadata>`
```
cbullinger, line 123:
```suggestion
To learn more, see :ref:`Authentication Provider Metadata <auth-provider-metadata>`.
```

cbullinger: not sure what the deal is, but when I click the link from the page, it takes me to the Enable User Data heading, but if I paste in the same URL, I get to the Authentication Provider Metadata heading 🤷‍♀️

cbullinger (approved):

source/authentication.txt
nlarew: Not sure what's happening either but probably a transient staging bug. Nothing in the source would make it point there so here's hoping it goes away in prod :)

PR #382, opened 2023-03-06
Add info about Email/Password Auth

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-20946

### Staged Changes

https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/docsp-27595/authentication/email-password/

### NOTE

The original ticket was looking for something closer to code guidance, which we could add to each SDK. I've taken the approach of making the conceptual page more clear.


source/authentication/email-password.txt
```diff
@@ -20,6 +20,20 @@ Overview
 Email/password authentication lets users register and login using an email address.
 Atlas App Services must confirm Email/Password users before they may log in.
 
+Users may choose one of three ways to handle user confirmation:
```
dacharyc, line 23: Right above this, we're talking about "Users" in the context of end users of an app registering and logging in, but here I think you're talking about developers, so I would avoid "Users" here.

```suggestion
Developers may choose one of three ways to handle user confirmation:
```

dacharyc: After reading the page, I think having this "three ways to handle user confirmation" section in the Overview is a little strange. I would expect to find this down below in the "User Confirmation" section. (And I see that a summary _is_ there as I would expect.) I think of the Overview as more of a high-level description and the three ways of handling user confirmation seem like implementation details that have been sort of shoe-horned in here.

source/authentication/email-password.txt
```diff
@@ -20,6 +20,20 @@ Overview
...
+
+- :ref:`Send a Confirmation Email <auth-send-a-confirmation-email>`
+
+- :ref:`Run a Confirmation Function <auth-run-a-confirmation-function>`
+
+- :ref:`Automatically Confirm Users <auth-automatically-confirm-users>`
+
+The following diagram shows the process flow for logging in:
+
+.. figure:: /images/email-password-login-flow.png
```
dacharyc, line 33: Love this diagram! But now I want relevant diagrams for _all_ the authentication providers. 🤔

source/authentication/email-password.txt
```diff
@@ -93,6 +94,22 @@ Configuration
 
 User Confirmation
 -----------------
+User confirmation occurs when a user wants to create a new Email/password
```
dacharyc, line 97: I wonder if we want to reframe this in terms of what the developer must do here? i.e. something like:

> Before an email/password user can authenticate, your app must register and 
> confirm the user. Registering an email/password user creates a new user object. 
> App Services then requires this user to confirm the creation of this account. You
> can select one of three methods to confirm a new user account:

source/authentication/email-password.txt
```diff
@@ -93,6 +94,22 @@ Configuration
 
 User Confirmation
 -----------------
+User confirmation occurs when a user wants to create a new Email/password 
+account. Atlas App Services supports three methods for confirming a user account:
+
+- Send a confirmation email
+- Run a confirmation function
+- Automatically create the account with no validation.
```
dacharyc, line 102: For consistency, we should either use punctuation for all or none of these bullet points.

```suggestion
- Automatically create the account with no validation
```

source/authentication/email-password.txt
```diff
@@ -93,6 +94,22 @@ Configuration
...
+
+.. warning:: 
+
+   Automatically creating accounts is useful when developing and testing an app, 
+   but you should not use it in a production app.
```
dacharyc, line 107: I wonder if we want to word this more strongly? Maybe something like: 
```suggestion
   Only use automatic confirmation when you are developing your app.
   Production applications should always use a secure confirmation process.
```

source/tutorial/dotnet.txt
```diff
@@ -74,23 +74,61 @@ Prerequisites
...
 
-.. code-block:: shell
+.. include:: /includes/tutorial-learn-more-template-app.rst
 
-   realm-cli apps create -n MyTutorialApp --template maui.todo.flex \
-      --deployment-model global --location us-or --environment development
+.. tabs::
 
-.. include:: /includes/tutorial-learn-more-template-app.rst
+   .. tab:: App Services UI
```
dacharyc: Oops - I think I saw that this update already got merged last week. Maybe rebasing from master will remove this file from this PR?

dacharyc (approved):
I love the diagrams! I have some Qs and suggestions for wording changes to frame this more about what the developer must do here and in the SDK. 

I think that these are some nice improvements to the page, but I'm not sure it directly corresponds to the request made in the Jira ticket. I wonder if we want to close the existing Jira ticket, but make related tickets (using Jira/related) for each SDK to beef up their documentation around managing email/password users?

source/authentication/email-password.txt
MongoCaleb: Slightly reworded because AAS doesn't necessarily require the _user_ to confirm their account. 

> Before an email/password user can authenticate, your app must register and
confirm the user's new account. Registering an email/password user creates a 
new user object. App Services then requires confirmation of the account. You can 
select one of three methods to confirm a new user account:

source/authentication/email-password.txt
```diff
@@ -212,30 +217,38 @@ The following table describes the potential values of this field:
 
    * - ``pending``
 
-     - App Services changes the user's confirmation status to ``pending``.
+     - App Services changes the user's confirmation status to ``Pending Confirmation``.
```
dacharyc, line 220: I wonder if we want to add something about the `Pending` case relative to the email being registered, as in attempting to re-register the user while the confirmation is `Pending` will fail because the user's email is in the system, even though their account hasn't been confirmed. I've seen people get tripped up on that before.

Maybe we want to point them to the [Admin API endpoint where they can list pending users](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/docsp-27595/admin/api/v3/#tag/users/operation/adminListPendingUsers), to check for a pending user account in that case?

Technically, I suppose that applies to both "Send a user confirmation email" and "confirm with a custom function" so maybe there should be a separate section at the top level of the User Confirmation section describing the three possible states, implications for the email being "in use", and how a dev can use the Admin API to check for a pending user?

source/authentication/email-password.txt
```diff
@@ -244,6 +257,11 @@ A user confirmation function is generally structured as follows:
       }
    }
 
+Unless the custom function automatically confirms the user, you must provide a
```
dacharyc, line 260: I think "you must provide a means for users to respond to the confirmation message that the function sends" is specific to this example, but I could see a developer interpreting it as "a confirmation function sends a message that the user must respond to" which isn't true inherently. It could be doing anything. Maybe we say instead something like:

> Unless the custom function automatically confirms the user, you must provide a
> means for the user to complete the confirmation process after the function fires.
> After completing any additional confirmation steps, call the SDK's Confirm User 
> method to finalize the user account creation on Atlas App Services.

dacharyc (commented):
We're getting there! I think we could stand to tweak the wording in one spot, and maybe provide some additional high-level information about the different user confirmation states common to both the custom function confirmation and send a user confirmation email flows.

source/authentication/email-password.txt
MongoCaleb: much better!

PR #383, opened 2023-03-06
(DOCSP-20646): Clarify $match example and add test

cbush:
## Pull Request Info

Ultimately this just clarifies the match expression docs, adds another example. It also adds a test to the test suite demonstrating the example, but it's not really used in the docs. It's just for our understanding.

Add functions test GitHub Action.

### Jira

- https://jira.mongodb.org/browse/DOCSP-20646

### Staged Changes

- https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/functions-match/triggers/database-triggers/#use-match-expressions-to-limit-trigger-invocations

### Reminder Checklist

### Review Guidelines


examples/functions/functions_tester/functions/messageHandler.js
```diff
@@ -0,0 +1,10 @@
+exports = function(changeEvent) {
+    const {fullDocument} = changeEvent;
+    
+    // ... Process the document here ...
+    const someResult = "approved";
+    
+    // Update the document with the result
+    const collection = context.services.get("mongodb-atlas").db("triggerExample").collection("messages");
+    return collection.updateOne({ _id: fullDocument._id, needsTriggerResponse: false, result: someResult });
```
mongodben, line 9: this returns a `Promise`. consider refactoring function to use `async` and this line to use `await`?

examples/functions/functions_tester/test/integration/triggerMessage.test.js
```diff
@@ -0,0 +1,50 @@
...
+    const collection = db.collection("messages");
+    const _id = BSON.ObjectId();
+    await collection.insertOne({
+      _id,
+      needsTriggerResponse: true,
+      result: null,
+    });
+
+    // Trigger should fire within the next 5 seconds
+    for (let attempt = 0; attempt < 10; ++attempt) {
```
mongodben, line 30: nit: standard to postfix ++ unless reason otherwise

```suggestion
    for (let attempt = 0; attempt < 10; attempt++) {
```

.github/workflows/test-functions.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
+      - run: cd examples/functions/functions_tester && npm ci && npm run test -- --forceExit
```
mongodben, line 16:
```suggestion
      - run: cd examples/functions/functions_tester && npm ci && npm run test --forceExit
```

pretty sure you don't need the `--` here. you only need that if you want to specify a path to files.

mongodben (approved):
lgtm. thx for adding. a few small things throughout

.github/workflows/test-functions.yml
cbush: As discussed on slack, the `--` is needed to denote passing the subsequent args directly to the inner command (`jest`) as opposed to the outer command (`npm run`)

examples/functions/functions_tester/test/integration/triggerMessage.test.js
cbush: That's not any standard I have ever heard of or would subscribe to. There is a semantic difference.

examples/functions/functions_tester/functions/messageHandler.js
cbush: I might agree if this were being used as an example (because people might not know they can just use the async keyword on the function), but returning `await foo` is basically the same as returning the promise, so I'm not sure I agree with you here?

mongodben: to clarify my thought, it'd be refactoring this to be: 

```js
// ..rest of code
const updateResult = await collection.updateOne({ _id: fullDocument._id, needsTriggerResponse: false, result: someResult });
return updateResult;
```

examples/functions/functions_tester/test/integration/triggerMessage.test.js
cbush: Yeah, but it's always been the wrong way to do it. You don't need the reference to the value before the increment, so there's no reason to use postfix. Some people even argue that either increment operator should be abolished and one should only use `+=`. I don't get it.

examples/functions/functions_tester/functions/messageHandler.js
cbush: I can definitely see where you're coming from

PR #384, opened 2023-03-06
Add flowcharts

cbush:
https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/rules-graph/rules/roles


source/rules/roles.txt
```diff
@@ -75,6 +75,24 @@ You can define field-level permissions for specific fields and
 default read/write permissions for any additional fields that you 
 don't explicitly define.
 
+Read Permissions Flowchart
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following diagram shows how App Services determines whether a user can
+read a given document:
+
+.. image:: /images/rules-read.png
```
dacharyc, line 84: It's pretty difficult to read the text on these charts without being able to expand them. Caleb put up a PR recently (#382) where you can "Click to Enlarge" one of the diagrams. The only difference I can see in the rST is that his PR has a `.. figure::` instead of an `.. image::` - would that enable us to click to expand these? If so, I'd recommend making these all figures.

dacharyc (approved):
I think it's fine, but I'd like to see the option to expand these diagrams because I find the flowchart text difficult to read. It's pretty small on my screen.

Big-picture, I would love to see us explore text-based alternatives to diagrams like this that might work with screen readers. (Maybe in addition to the images? Maybe something like [mermaid](https://github.com/mermaid-js/mermaid)?)

PR #391, opened 2023-03-14
(DOCSP-27347): Add reset email endpoint to Admin API

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-27347

### Staged Changes

- [Admin API](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-27347/admin/api/v3/#tag/users/operation/adminResetUserEmail)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -1698,12 +1698,11 @@ paths:
         "204":
           description: No Content
         "400":
-          description:
-            oneOf:
-              - Payload is formatted incorrectly.
-              - The email in the payload is not a string.
-              - The email in payload has 0 or greater than 128 characters.
-              - The user does not have a local-userpass identity.
+          description: |-
+            - Payload is formatted incorrectly.
+            - The email in the payload is not a string.
+            - The email in payload has 0 or greater than 128 characters.
+            - The user does not have a local-userpass identity.
         "404":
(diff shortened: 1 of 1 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -1670,6 +1670,42 @@ paths:
...
+            example: { "email": "new_email@example.com" }
+      responses:
+        "204":
+          description: No Content
+        "400":
+          description: |-
+            - Payload is formatted incorrectly.
+            - The email in the payload is not a string.
+            - The email in payload has 0 or greater than 128 characters.
+            - The user does not have a local-userpass identity.
```
nlarew, line 1705: Huh, I'm a urprised this isn't a 404.

nlarew (approved):

source/openapi-admin-v3.yaml
dacharyc: I actually meant to ask you - these are listed in the spec as 400s, but I didn't actually get this wording for any of these conditions. I thought it was useful to list these as reasons you could get a 400, but I was wondering if this "description" field is meant to be only the string text that accompanies the status (which was something like malformed payload, IIRC), or if I can use it like this to list the potential causes for the status.

PR #394, opened 2023-03-15
(DOCSP-27081) [UI] Configure user session refresh token expiration

nlarew:
### Jira

- (DOCSP-27081) [UI] Configure user session refresh token expiration

### Staged Changes

- [Manage User Sessions > Configure Refresh Token Expiration](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/refresh-ui/users/sessions/#std-label-refresh-token-expiration)

### DIFF
--- a/source/users/sessions.txt
+++ b/source/users/sessions.txt
@@ -316,15 +316,35 @@ configure this time for your App's refresh tokens to be anywhere between
 - Custom JWT refresh token expiration is determined by the ``exp`` value
   of the user's JWT.
 
-.. note::
+You can configure the refresh token expiration time for all sessions in
+an App from the Admin UI or Admin API.
 
-   You can currently only configure refresh token expiration with the
-   Admin API.
+Set the Refresh Token Expiration In the Admin UI
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


source/users/sessions.txt
```diff
@@ -316,15 +318,35 @@ configure this time for your App's refresh tokens to be anywhere between
 - Custom JWT refresh token expiration is determined by the ``exp`` value
   of the user's JWT.
 
-.. note::
+You can configure the refresh token expiration time for all sessions in
+an App from the Admin UI or Admin API.
+
+Set the Refresh Token Expiration In the Admin UI
```
dacharyc, line 324: Nit: I don't think we generally refer to an "Admin UI" in our docs? I just did a little clicking around and I see us calling it "App Services UI" in tab sets and other places. I might suggest renaming this either here or elsewhere for consistency.

dacharyc (approved):
Overall, LGTM - just a question about the use of "Admin UI" vs. "App Services UI."

source/users/sessions.txt
nlarew: We introduced "The App Services Admin UI" as the full name in our `/app` section IA refresh so it's limited to that section atm. I think if we use it more it has nice parallels since it's basically a UI for the Admin API.

PR #396, opened 2023-03-16
(DOCSP-27564) Add documentation for sync migrations

krollins-mdb:
## Pull Request Info

Waiting on required `realm-core` and SDK versions. Everything else is ready for review.

### Jira

- https://jira.mongodb.org/browse/DOCSP-27564

### Staged Changes

- [Migrate Partiion-Based Sync to Flexible Sync](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-27564/sync/migrate-sync-modes/)

### Reminder Checklist

### Review Guidelines


source/reference/partition-based-sync.txt
```diff
@@ -1020,7 +1020,18 @@ cannot express solely in JSON expressions.
        collection and then returns a boolean that indicates if the user can read
        data in the requested partition.
 
+.. _pbs-rule-migration:
 
+Migrate Partition-Based Sync Rules to Flexible Sync Rules
```
kmorkos, line 1025: now that flexible sync uses app services rules, I believe we should just refer to this as "App Services Rules" (I'm actually not sure what term is used in other parts of the docs)

Some links that may be relevant: 
- https://www.mongodb.com/docs/atlas/app-services/rules/roles/#with-device-sync
- https://www.mongodb.com/docs/atlas/app-services/sync/app-builder/device-sync-permissions-guide/#std-label-flexible-sync-permissions-guide

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+   :class: singlecol
+
+You can automatically migrate App Services Apps that use Partition-Based Sync
+to Flexible Sync. We recommend doing this, as Partition-Based Sync is an older
+mode.
+
+Other than upgrading the SDK version, migrating sync modes doesn't require any 
+changes to your client app code. Your App's backend will automatically translate
+the client's Partition-Based Sync requests into Flexible Sync.
+.. QUESTION: is there a performance hit for the translation?
```
kmorkos, line 20: nope! there is a one-time cost (roughly equivalent to the regular cost of a client reset) when the PBS client first tries to connect, but once that handshake happens it will be a regular flexible sync client moving forward

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+------------
+
+- Your App must currently use Partition-Based Sync
+- Linked Atlas cluster must be running mongodb 5.0+ for Flexible Sync compatibility
+- Linked Atlas cluster must have time-based oplog setting enabled
+- realm-core vTBD+, which means these SDK versions: TBD
+
+.. _realm-sync-migrate-what-to-know:
+
+What to Know Before migrating
```
kmorkos, line 36: another important callout here is that old clients will get a client reset, so they should make sure they have a client reset handler mode that makes sense for them (and maybe worth encouraging them to use client reset with recovery if they don't already)

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+  period of downtime.
+- If Partition-Based Sync default client max offline time is unset, the new
+  Flexible Sync config will set it to 30 days.
+- Storage in your Atlas cluster used by sync is temporarily doubled. Partition-Based
+  Sync and Flexible Sync must temporarily co-exist. Depending on the size of your
+  synced data, this could affect billing.
+
+.. QUESTION: If someone were to cancel a migration - or roll back a migration,
+   they'd potentially be on the hook for costs associated with subsequent 
+   migration attempts, correct?
```
kmorkos, line 98: what costs are you referring to here?

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+Your App will lose the migration in the following cases:
+
+- Partition-Based Sync is re-enabled
+- You roll back the migration
+
+Migrate permissions
+```````````````````
+
+Your App's sync permissions will be automatically migrated if the Partition-Based
+Sync permissions have direct Flexible Sync equivalents.
```
kmorkos, line 114: Worth mentioning that in this case, it will override any app services rules that were previously defined

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+	- Users should test these and check if they need to make changes to the auto-migrated rules.
+	- Testing new permissions won't work on existing app. Need a new app to test manually-migrated permissions rules.
+
+Refer to the :ref:`Partiton-Based Sync <pbs-rule-migration>` reference for examples
+of how to map these Partition-Based Sync permissions to Flexible Sync permissions.
+
+Commit Migration
+~~~~~~~~~~~~~~~~
+
+After your migration has completed, it is in a non-permanent state. You can manually
```
kmorkos, line 134: [nit] not a huge fan of "non-permanent" here since it may imply that we would roll it back without manual intervention

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+commit the migration to make it permanent or roll back the migration to return your
+App to Partiton-Based Sync.
+
+If you don't commit or roll back the migration, it will automatically commit when
+the PBS translator falls off of the oplog.
+.. QUESTION: What will the oplog show, exactly, for PBS translator
```
kmorkos, line 140: This is directly proportional to the cluster's time-based oplog setting: https://www.mongodb.com/docs/atlas/cluster-additional-settings/#set-minimum-oplog-window

e.g. if they set the value to 24 hours, then they can assume that they have ~24 hours to roll back starting from when the migration _started_ (not ended!)

while the rollback window is open, they can continue to bump that value to expand their window at the cost of additional cluster storage

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+
+You can roll back a migration after it has completed and before it has been committed.
+
+Completed migrations are automatically committed when the PBS translator falls 
+off of the oplog.
+.. QUESTION: What will the oplog show, exactly, for PBS translator
+
+You can increase or decrease the migration roll back window by adjusting
+oplog retentention.
+.. TODO: Add link to oplog retention docs
```
kmorkos, line 180: here ya go - https://www.mongodb.com/docs/atlas/cluster-additional-settings/#set-minimum-oplog-window

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+Migrating results in Partition-Based Sync clients communicating with a
+Flexible Sync App Services App. You cannot reverse this and have Flexible Sync
+clients communicate with a Partition-Based Syn App Services App.
+
+Requirements
+------------
+
+- Your App must currently use Partition-Based Sync
+- Linked Atlas cluster must be running mongodb 5.0+ for Flexible Sync compatibility
+- Linked Atlas cluster must have time-based oplog setting enabled
```
kmorkos, line 31: this carries with it two other implicit requirements:
- linked cluster must have storage auto-scaling enabled
- linked cluster must be M10+ to access this setting

source/sync/migrate-sync-modes.txt
krollins-mdb: Ah, sorry. Potential costs related to the doubling sync metadata. It's doubled every time because they have to start over, right?

source/reference/partition-based-sync.txt
krollins-mdb: Ah, yes! Thanks for pointing that out. I'll refactor throughout.

krollins-mdb: We only use "App Services Rules" in a couple places in the docs. I think we default to "rules" being implicitly associated with App Services Rules. I'll double check with the team though.

Either way, I think it makes sense to call it App Services Rules here.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,192 @@
...
+Begin Migration
+~~~~~~~~~~~~~~~
+
+Before you begin a migration, make sure you're aware of the
+:ref:`migration stages <realm-sync-migrate-stages>` and
+:ref:`effects <realm-sync-migrate-what-to-know>`.
+
+During migration, client apps may experience side effects, most of which are similar
+to what happens if you disable and re-enable sync. These are some side effects
+of migration:
```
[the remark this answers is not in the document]
krollins-mdb: Great point! I'll clarify that these are additional side effects.

source/sync/migrate-sync-modes.txt
krollins-mdb: Yeah, I don't think it's quite right. I've updated to `an evaluation state`. Do you think that conveys the purpose of the roll back window?

source/sync/migrate-sync-modes.txt
krollins-mdb: Ooh, those are good details. I remember you mentioning auto-scaling, but I didn't know you need M10+ to access it.

I'll add these to make them clear to readers.

source/sync/migrate-sync-modes.txt
kmorkos: Ah I see. Canceling/rolling back a migration automatically clears the flexible sync metadata - so their storage should go back to what it was prior to starting the migration

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,245 @@
...
+Linked Atlas Cluster:
+
+- Must be running mongodb 5.0+ for Flexible Sync compatibility
+- Must have storage auto-scaling enabled
+- Must be M10+ to access this setting
+- Must have time-based oplog setting enabled
+
+Client Apps:
+
+- ``realm-core`` vTBD+, which means these SDK versions: TBD
```
krollins-mdb, line 41: @kmorkos, do we know what version of realm-core and SDK versions users will need?

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,245 @@
...
+There are three stages to a Sync mode migration. You can see your migration's
+stage in the App Services UI.
+
+#. Disabling Partition-Based Sync
+   
+   At this stage, client apps can connect to your App's backend, but local writes 
+   will not sync. These writes will be lost if you don't have a 
+   :ref:`client reset handler <handle-a-client-reset>` to recover unsynced changes.
+	 
+   Clients can also receive updates to MongoDB data, but they won't
```
krollins-mdb, line 79: @kmorkos, "MongoDB data" in this instance refers to the linked cluster, correct?

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,245 @@
...
+
+Before you begin a migration, make sure you're aware of the
+:ref:`migration stages <realm-sync-migrate-stages>` and
+:ref:`effects <realm-sync-migrate-what-to-know>`.
+
+During migration, client apps may experience side effects, most of which are similar
+to what happens if you disable and re-enable sync. These are some additional
+side effects of migration:
+
+- In the App Services UI, the Rules and Schemas pages will be unresponsive.
```
krollins-mdb, line 112: @kmorkos, do you think "unresponsive" is a fair description of what will happen? I haven't seen that side effect.

source/sync/migrate-sync-modes.txt
kmorkos: Not yet, tagging @michael-wb since he's currently working on the realm-core side

source/sync/migrate-sync-modes.txt
kmorkos: correct

source/sync/migrate-sync-modes.txt
kmorkos: It wouldn't be unresponsive, just read-only. So they can still see their schemas (I think there's a bug that might lead to a broken view of rules), but if they try to make any changes to their config it will fail

krollins-mdb: Ah, that makes a lot more sense. Thank you!

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,253 @@
...
+:ref:`migration stages <realm-sync-migrate-stages>` and
+:ref:`effects <realm-sync-migrate-what-to-know>`.
+
+During migration, client apps may experience side effects, most of which are similar
+to what happens if you :ref:`terminate <terminating-realm-sync>` and re-enable
+Device Sync. These are some additional side effects of migration:
+
+- In the App Services UI, the Rules and Schema pages will be read-only.
+- Your App will experience a period of read-only sync and then a 
+  period of downtime.
```
cbush, line 110: Do we know how long the downtime might be?

cbush (approved):

source/sync/migrate-sync-modes.txt
krollins-mdb: Unfortunately, no. And it's hard to give an estimate without potentially putting ourselves in a difficult situation when we're inevitably wrong. I could repeat what I have in the "What to Know Before Migrating" section:
> How long a migration takes is directly proportional to the amount of data that
   needs to be migrated. The more data, the longer it will take to migrate.

Do you think that would be beneficial?

krollins-mdb:
@kmorkos, after talking with @cbush, we decided to be more generic with the rules migration info - pointing readers to existing information. If we know that there's a specific rule migration that a significant number of readers would struggle with, we can look at including it. Otherwise, there are so many variables, it's hard to know what's of real benefit.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,253 @@
...
+- Must be M10+ to access this setting
+- Must have time-based oplog setting enabled
+
+Client apps:
+
+- ``realm-core`` vTBD+, which means these SDK versions: TBD
+
+.. _realm-sync-migrate-what-to-know:
+
+What to Know Before migrating
```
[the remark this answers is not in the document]
krollins-mdb: Applied locally with other small edits. Will be in next push.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,253 @@
...
+   :backlinks: none
+   :depth: 3
+   :class: singlecol
+
+Atlas App Services Apps can have one of two Device Sync modes: Partition-Based Sync
+and Flexible Sync. Partition-Based Sync is an older mode and you should consider
+migrating to Flexible Sync.
+
+Migrating an App that uses Partition-Based Sync to Flexible Sync is an automatic
+process. Other than upgrading the SDK version, migrating sync modes doesn't require
```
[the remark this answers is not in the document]
krollins-mdb: Applied locally with other small edits. Will be in next push.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,251 @@
...
+previously-defined App Services Rules.
+
+Some Partition-Based Sync rule strategies can't translate directly to App Services
+Rules. You may need to manually migrate rules with these expansions:
+
+- Rules that include :json-expansion:`%%partition`. 
+  
+  App Services Rules do not have an equivalent expansion.
+  :ref:`Function rules <pbs-function-rules>` used with :json-expansion:`%%partition`
+  cannot be migrated.
```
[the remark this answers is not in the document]
krollins-mdb: Oh, I misunderstood. I thought `%function` would work, but not in conjunction with `%%partition`. Thank you for the clarification. I'll update this section to account for that.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,251 @@
...
+   Partition-Based Sync Metadata.
+
+.. _realm-sync-migrate-revert:
+
+Revert Migration
+~~~~~~~~~~~~~~~~
+
+You can revert a migration after it has completed and before it has been committed.
+
+Completed migrations are automatically committed when the PBS translator falls
```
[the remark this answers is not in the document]
krollins-mdb: Thank you! I thought I fixed that, but clearly left it in. We do not abbreviate to PBS in the docs.

kmorkos:
> @kmorkos, after talking with @cbush, we decided to be more generic with the rules migration info - pointing readers to existing information. If we know that there's a specific rule migration that a significant number of readers would struggle with, we can look at including it. Otherwise, there are so many variables, it's hard to know what's of real benefit.

Makes sense to me - thanks for following up!

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,251 @@
...
+
+During migration, client apps may experience side effects, most of which are similar
+to what happens if you :ref:`terminate <terminating-realm-sync>` and re-enable
+Device Sync. These are some additional side effects of migration:
+
+- In the App Services UI, the Rules and Schema pages will be read-only.
+- Your App will experience a period of read-only Sync and then a 
+  period of downtime.
+- If Partition-Based Sync :ref:`client max offline time <client-maximum-offline-time>`
+  is unset, the new Flexible Sync configuration will set it to 30 days.
```
kmorkos, line 112: UPDATE: we're removing the "client max offline time" setting for PBS apps, so we decided that for a migration, we will always set the flexible sync max offline time to 30 days regardless of their old PBS setting

source/images/commit-migration.png
kmorkos, line 1: FYI - you can get the updated screenshots from the description in https://github.com/10gen/baas-ui/pull/3093

I'll let you know when that's merged so we know the copy is final

kmorkos: It's merged! Will hold off on taking a final pass until the screenshots are updated

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,271 @@
...
+  needs to be migrated. The more data, the longer it will take to migrate.
+- Migrations can be :ref:`canceled <realm-sync-migrate-cancel>` during the 
+  migration or :ref:`reverted <realm-sync-migrate-revert>` within a certain 
+  timeframe after the migration is finished.
+- Progress notifications will not work after migration because Flexible Sync does
+  not support them.
+- Client apps that use an SDK with with a realm-core version lower than TBD will
+  experience a client reset after migration. You should use a 
+  :ref:`client reset <handle-a-client-reset>` handler with recovery enabled.
+- :ref:`Development Mode <development-mode>` is not supported during a migration.
```
kmorkos, line 61: This is a bit misleading - dev mode can still be enabled on the backend, and client apps can still add new columns to existing tables without issues. The scenario that @michael-wb and @danieltabacaru were saying would not work properly is if a client tried to create a new table against an FLX backend while using a PBS config locally.

I don't think it's worth getting into all of the nuances about what exact behavior they will see, so maybe something simple like `Creating new tables via Development Mode may lead to unexpected behavior while using a partition-based sync configuration in your app`

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,271 @@
...
+Your App's permissions and rules will be automatically migrated if the Partition-Based
+Sync rules have direct App Services Rules equivalents. This will override
+previously-defined App Services Rules.
+
+In step 4 above, you will see if your permissions and rules can be automatically
+migrated. On the :guilabel:`Review Migration Process` step in the App Services UI,
+look for :guilabel:`Rules & Permisssions`.
+
+Some Partition-Based Sync rule strategies can't translate directly to App Services
+Rules. You may need to manually migrate rules that include:
```
kmorkos, line 158: [nit] "rules" -> "permissions" might be more familiar for a PBS user

source/sync/migrate-sync-modes.txt
krollins-mdb: Ha! Yeah, my first local revision for this had a ton of detail. I decided to pare it back and went too far. Thanks.

kmorkos (approved):
Looks great once the realm-core version is filled in!

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,275 @@
...
+Linked Atlas Cluster:
+
+- Use MongoDB 5.0 or later for Flexible Sync compatibility
+- Have storage auto-scaling enabled
+- Be an M10+ Cluster Tier
+- Have time-based oplog setting enabled
+
+Client apps:
+
+- Use ``realm-core`` vTBD+, which means these SDK versions: TBD
```
krollins-mdb, line 41: `realm-core` version possibly 13.8.0.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,328 @@
...
+Migrate Client App to Flexible Sync
+-----------------------------------
+
+After your app's backend has been migrated to Flexible Sync, we recommend
+that you also migrate your client app to Flexible Sync. When you migrate your
+client app to Flexible Sync, the client will stop automatically creating
+sync subscriptions for each Realm object model in your app's data model.
+
+We also recommend removing all sync subscriptions and then creating new subscriptions
+for your data. This is the clearest way to sync data after a migration.
```
kmorkos, line 299: hmm, not sure i get what this second sentence means. IIUC, maybe "This is the clearest way to control what data is synced after a migration" ?

kmorkos (commented):
> @kmorkos, I made some additions based on the new information you shared. What do you think? I'm considering this a "first pass" on the new info and would like to know if I'm going in the right direction.

@krollins-mdb looks great so far! I left a few minor comments

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,328 @@
...
+#. **Completed migration** 
+
+   :guilabel:`Manage Sync Migration` appears on the Device Sync page if the 
+   migration hasn't been committed or reverted.
+
+After your backend is migrated to Flexible Sync, the backend automatically
+injects a ``partitionKey`` field into every document the Partition-Based Sync
+client creates and sets its value to your sync configuration's
+``partitionValue``. The injected ``partitionKey`` field does not sync back to
+client devices. This means that clients won't be able to read this field.
```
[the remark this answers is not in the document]
krollins-mdb: Ooh, that's a good thing to clarify. Thank you!

source/sync/migrate-sync-modes.txt
krollins-mdb: Yes, thank you for pointing out the ambiguity here.

source/reference/partition-based-sync.txt
```diff
@@ -1014,7 +1016,39 @@ cannot express solely in JSON expressions.
...
+  and cannot be migrated.
+
+- :json-expansion:`%%partition` expansion.
+
+  App Services Rules do not have an equivalent expansion for  
+  :json-expansion:`%%partition`, so it cannot be migrated.
+
+- ``%or``, ``%not``, ``%nor``, ``%and`` expansions.
+
+  These permissions may work, but there is enough nuance that you should test them
```
MongoCaleb, line 1042: nit: consider making "may" italics.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,348 @@
...
+   :backlinks: none
+   :depth: 3
+   :class: singlecol
+
+Atlas App Services Apps can have one of two Device Sync modes: Partition-Based Sync
+and Flexible Sync. Partition-Based Sync is an older mode and you should consider
+migrating to Flexible Sync.
+
+Migrating an App that uses Partition-Based Sync to Flexible Sync is an automatic
+process. Other than upgrading the SDK version, migrating Sync modes doesn't require
```
MongoCaleb, line 18: Is upgrading the SDK required? Seems like only if it is old-ish. As I read the sentence, it implies one must upgrade. If not, perhaps: Other than possibly needing to upgrade the SDK version...

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,348 @@
...
+   :local:
+   :backlinks: none
+   :depth: 3
+   :class: singlecol
+
+Atlas App Services Apps can have one of two Device Sync modes: Partition-Based Sync
+and Flexible Sync. Partition-Based Sync is an older mode and you should consider
+migrating to Flexible Sync.
+
+Migrating an App that uses Partition-Based Sync to Flexible Sync is an automatic
```
MongoCaleb, line 17: I think "app" should be lowercase here.

MongoCaleb: And are we talking about a client app or AAS app?

source/sync/migrate-sync-modes.txt
MongoCaleb: Update before deploying!

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,348 @@
...
+Client apps:
+
+- Use ``realm-core`` vTBD+, which means these SDK versions: TBD
+
+.. _realm-sync-migrate-what-to-know:
+
+What to Know Before Migrating
+-----------------------------
+
+Migrating your App's Sync mode involves changes that affect both client apps and
```
MongoCaleb, line 49: App => app throughout

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,348 @@
...
+your App's backend. You should be aware of the side effects and plan ahead.
+
+- Client apps that use an SDK with with a realm-core version lower than TBD will
```
MongoCaleb, line 52: Watch out for the "TBD"s!!!

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,348 @@
...
+- Creating new tables in a migrated Flexible Sync backend using 
+  :ref:`Development Mode <development-mode>` from a Partition-Based Sync client
+  may lead to unexpected behavior.
+
+.. _realm-sync-migrate-stages:
+
+Migration Stages
+~~~~~~~~~~~~~~~~
+
+There are three stages to a Sync mode migration. You can see your migration's
```
MongoCaleb, line 73: Consider switching the order of these first two sentences. That way, the second one is a lead-in to the list that follows. Might even consider ending it with a colon (:)

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,348 @@
...
+or manually commit the migration.
+
+Begin Migration
+~~~~~~~~~~~~~~~
+
+Before you begin a migration, make sure you're aware of the
+:ref:`migration stages <realm-sync-migrate-stages>` and
+:ref:`effects <realm-sync-migrate-what-to-know>`.
+
+During migration, client apps may experience side effects, most of which are similar
```
MongoCaleb, line 115: "Side effects may include painful rash..." ;-) 
Seriously, I always think of "side effects" in that way, and I wonder if others do, too? Also, do client apps "experience" anything? (not to be too philosophical here...)

If so, consider "client apps may be impacted similar to what happens when you terminate (etc)"

source/sync/migrate-sync-modes.txt
MongoCaleb: I see you added that note above. I don't think it needs to be in two places on the same page.

MongoCaleb (commented):
A few things to consider

source/sync/migrate-sync-modes.txt
```diff
@@ -285,6 +279,31 @@ start over from the beginning with future migration attempts.
...
 
+.. _realm-sync-migrate-after:
+
+After Backend Migration
+-----------------------
+
+Your migrated backend uses a partition key field in your Realm object models
+to map the Partition-Based Sync client objects to Flexible Sync equivalents on
+the backend. If your object models don't have a partition key field, the
+backend automatically injects one into every table the Partition-Based Sync
```
kmorkos, line 290:
```suggestion
backend automatically injects one into every object the Partition-Based Sync
```

source/sync/migrate-sync-modes.txt
```diff
@@ -285,6 +279,31 @@ start over from the beginning with future migration attempts.
...
+client creates.
+
+The injected field follows this structure: ``<yourPartitionKey> == 
+<yourPartitionValue>``. Both the field key and value are derived from your 
+Partition-Based Sync client configuration.
```
kmorkos, line 295: [nit] not sure if this is a distinction worth making, but technically `yourPartitionKey` is a backend parameter, while `yourPartitionValue` is a client config parameter

source/sync/migrate-sync-modes.txt
```diff
@@ -285,6 +279,31 @@ start over from the beginning with future migration attempts.
...
+
+There are some other characteristics of the injected field you should be aware of:
+
+- The injected field does not sync back to the client that created the table.
```
kmorkos, line 299:
```suggestion
- The injected field does not sync back to the client that created the object.
```

source/sync/migrate-sync-modes.txt
```diff
@@ -285,6 +279,31 @@ start over from the beginning with future migration attempts.
...
+  This means the originating client won't be able to read the injected
+  ``partitionKey`` field.
+- The Partition-Based Sync client automatically creates sync subscriptions for
```
kmorkos, line 302: this is not really a characteristic of the injected field, so it should probably be pulled out of this list

kmorkos (approved):
few final comments!

source/sync/migrate-sync-modes.txt
krollins-mdb: Ah, good point. I'll make this clearer.

source/sync/migrate-sync-modes.txt
krollins-mdb: It is required. There's some automatic logic added to the SDK to enable the PBS client to communicate with the Flexible Sync backend. SDK versions older than the (TBD) version that includes the logic simply won't be able to communicate with the backend.

source/sync/migrate-sync-modes.txt
krollins-mdb: Thanks for the reminder! We still don't know what Core version (or SDK versions) will be required yet.

source/sync/migrate-sync-modes.txt
krollins-mdb: Those TBDs'll get ya every time. 😆 

I'll update when we've got better version info.

source/sync/migrate-sync-modes.txt
krollins-mdb: Excellent suggestion. Thank you!

source/sync/migrate-sync-modes.txt
krollins-mdb: Ah, there I go anthropomorphizing technology again. I do hope every realm has a beautiful experience though. 😄 

I'll re-word this.

source/sync/migrate-sync-modes.txt
krollins-mdb: Ah, yes. Thank you for noticing this discrepancy. I've re-organized a bit.

source/sync/migrate-sync-modes.txt
krollins-mdb: Ah, thanks for clarifying.
> worth clarifying that the field name will be the partition key from their old backend PBS configuration, not the literal field name "partitionKey"

I guess I missed the "backend" part of what you said earlier. Sorry! I'll update.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,349 @@
...
+Linked Atlas Cluster:
+
+- Use MongoDB 5.0 or later for Flexible Sync compatibility
+- Have storage auto-scaling enabled
+- Be an M10+ Cluster Tier
+- Have time-based oplog setting enabled
+
+Client apps:
+
+- Use ``realm-core`` vTBD+, which means these SDK versions: TBD
```
kmorkos, line 42:
```suggestion
- Use ``realm-core`` v13.10.0+, which means these SDK versions: TBD
```

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,349 @@
...
+
+.. _realm-sync-migrate-what-to-know:
+
+What to Know Before Migrating
+-----------------------------
+
+Migrating your App's Sync mode involves changes that affect both client apps and
+your App's backend. You should be aware of the effects and plan ahead.
+
+- Client apps that use an SDK with with a ``realm-core`` version lower than TBD will
```
kmorkos, line 52:
```suggestion
- Client apps that use an SDK with with a ``realm-core`` version lower than v13.10.0 will
```

kmorkos (commented):
We have a realm-core version!

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+Migration generally has similar effects as :ref:`terminating <terminating-realm-sync>` 
+and re-enabling Device Sync. There are some additional effects of migration:
+
+- In the App Services UI, the Rules and Schema pages will be read-only.
+- Your App will experience a period of read-only Sync and then a 
+  period of downtime.
+- The new Flexible Sync configuration will set 
+  :ref:`client max offline time <client-maximum-offline-time>` to 30 days.
+- Storage in your Atlas cluster used by Device Sync metadata is temporarily doubled.
+  Partition-Based Sync metadata and Flexible Sync metadata must temporarily co-exist.
```
dacharyc, line 131: Is there any information about _how long_ this metadata must temporarily co-exist? i.e. is it for the client max offline time or some other definitive period, or is it variable?

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+   :guilabel:`Next: Migration Requirements`.
+#. Ensure you meet the migration requirements, the select 
+   :guilabel:`Start Migration`.
+
+.. note:: Temporary Sync Disruption
+
+   Clients will not be able to sync their local writes until the migration has
+   completed.
+
+After a migration is completed, Partition-Based Sync client apps can communicate
```
dacharyc, line 153: You mentioned above that clients must perform a client reset after migrating. Maybe it's worth adding that here?

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+- :json-expansion:`%%partition` expansion.
+
+  App Services Rules do not have an equivalent expansion for  
+  :json-expansion:`%%partition`, so it cannot be migrated.
+
+- ``%or``, ``%not``, ``%nor``, ``%and`` expansions.
+
+  These permissions may work, but there is enough nuance that you should test them
+  to ensure expected behavior. Testing new permissions won't work on the App
+  you are migrating. You'll need a new app to test manually-migrated permissions.
```
dacharyc, line 189: There's a little ambiguity here. Are we advising people to test manually-migrated permissions in a new FS app, and _then_ apply them manually here so they can complete the migration?

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+to map the Partition-Based Sync client objects to Flexible Sync equivalents on
+the backend. If your object models don't have a partition key field, the
+backend automatically injects one into every object the Partition-Based Sync
+client creates.
+
+These are some other characteristics of the injected field you should be aware
+of:
+
+- The injected field follows this structure: ``<yourPartitionKey> == 
+  <yourPartitionValue>``. Both the field key is derived from your old
```
dacharyc, line 303:
```suggestion
  <yourPartitionValue>``. The field key is derived from your old
```

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+until you :ref:`migrate your client app <realm-sync-migrate-client>` to
+Flexible Sync.
+
+.. _realm-sync-migrate-client:
+
+Migrate Client App to Flexible Sync
+-----------------------------------
+
+After your app's backend has been migrated to Flexible Sync, we recommend
+that you also migrate your client app to Flexible Sync. When you migrate your
```
dacharyc, line 321: In the SDK side, we should spell out exactly what it means to "migrate your client app to Flexible Sync." I assume that's something like:

- Update a Partition-Based Configuration to a Flexible Sync configuration
- Keep, remove, or update the automatically-generated subscriptions as needed (the Core PR says subscriptions will be given a name in the format "flx_migrated_")
- Add any fields models need to manage subscriptions. i.e. "If your object models don't have a partition key field, the backend automatically injects one into every object the Partition-Based Sync client creates... The injected field does not sync back to the client that created the object." Would we advise devs to specifically add such a key to their models so they're visible and can be updated as needed? And so they remember to give new models the relevant key?

source/sync/migrate-sync-modes.txt
```diff
@@ -26,6 +26,10 @@ Partition-Based Sync.
 Requirements
 ------------
 
+Currently, the Kotlin and Swift SDKs can migrate from Partition-Based Sync to
+Flexible Sync. The C++, Flutter, Java, .NET, Node.js, and React Native SDKs
```
kmorkos, line 30: can remove flutter from this list since migration is irrelevant there. i'm also not sure if the c++ sdk supports PBS? can't find anything in the docs indicating that it does

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+   :ref:`Flexible Sync - Node.js SDK <node-flexible-sync>`
+
+- React Native SDK
+   :ref:`Flexible Sync - React Native SDK <react-native-flexible-sync>`
+
+- Swift SDK
+   :ref:`Manage Flexible Sync Subscriptions - Swift SDK <swift-manage-flexible-sync-subscriptions>`
+
+If you don't remove all subscriptions, the automatically-created subscriptions
+will remain, but you won't see them in your client code. This makes it harder
```
dacharyc, line 355: "...but you won't see them in your client code" was a little confusing to me. I assumed they would be available in the realm.subscriptions set with the autogenerated names along with any other subscriptions. But after re-reading this I think you mean something like "you don't see a code path in your code base that creates these subscriptions so you might forget these exist." (Which is probably also true of the partition key we automagically create.)

I don't have a specific suggestion, but I wonder if it's worth trying to make clearer that these subscriptions are accessible on the client but because they weren't created in the client codebase, it might make maintaining and extending client code confusing.

source/sync/migrate-sync-modes.txt
michael-wb: Correct - Flutter and C++ are the newest SDKs and only support flexible sync.

dacharyc (commented):
I don't think any of this feedback is blocking, but I had a few comments throughout.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+Linked Atlas Cluster:
+
+- Use MongoDB 5.0 or later for Flexible Sync compatibility
+- Have storage auto-scaling enabled
+- Be an M10+ Cluster Tier
+- Have time-based oplog setting enabled
+
+Client apps:
+
+- ``realm-core`` v13.10.0 or later, which is included in these SDK versions:
```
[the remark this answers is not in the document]
krollins-mdb: Good point. No need to make devs think about realm-core.

source/sync/migrate-sync-modes.txt
krollins-mdb: Good point. It only lasts as long as the migration itself. As soon as the migration is rolled back or completed, the doubled metadata is removed.

krollins-mdb: Actually, @kmorkos, what happens if there isn't enough space in the cluster to double the metadata? Would this be a rare edge case?

source/sync/migrate-sync-modes.txt
krollins-mdb: Thanks for catching this. "Both the field key" just rolls of the tongue though... 😆

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+your App's backend. You should be aware of the effects and plan ahead.
+
+- Client apps that use an SDK with with a ``realm-core`` version earlier than
+  v13.10.0 will not be able to sync after migration.
+- How long a migration takes is directly proportional to the amount of data that
+  needs to be migrated. The more data, the longer it will take to migrate.
+- Migrations can be :ref:`canceled <realm-sync-migrate-cancel>` during the 
+  migration or :ref:`reverted <realm-sync-migrate-revert>` within a certain 
+  timeframe after the migration is finished.
+- Progress notifications will not work after migration because Flexible Sync does
```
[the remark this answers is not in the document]
krollins-mdb: @kmorkos, would having client code for progress notifications cause fatal client errors? I know one of the goals of the project was to prevent devs from needing to make client code changes. Is this something that has to be done?

source/sync/migrate-sync-modes.txt
krollins-mdb: Yeah, this section was tricky to write. It's a "probably will work, but we can't guarantee it" situation. And, yes, your read is correct. I'll edit to make that clearer.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+
+After your app's backend has been migrated to Flexible Sync, we recommend
+that you also migrate your client app to Flexible Sync. When you migrate your
+client app to Flexible Sync, the client will stop automatically creating
+sync subscriptions for each Realm object model in your app's data model.
+
+We also recommend removing all sync subscriptions and then creating new subscriptions
+for your data. This is the clearest way to control what data is synced after a
+migration. Refer to the SDK documentation for details about adding and removing
+subscriptions.
```
michael-wb, line 333: I don't know if you want to mention the naming format of the generated subscriptions while the client is in the migrated state. It follows this format: `flx_migrated_` followed by the name of the table/collection. For example, a subscription for the "Person" collection will be named: `flx_migrated_Person`

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+.. _realm-sync-migrate-after:
+
+After Backend Migration
+-----------------------
+
+Your migrated backend uses a partition key field in your Realm object models
+to map the Partition-Based Sync client objects to Flexible Sync equivalents on
+the backend. If your object models don't have a partition key field, the
+backend automatically injects one into every object the Partition-Based Sync
+client creates.
```
michael-wb, line 302: You might want to point out that this value will no longer be injected once the client app has been migrated to flexible sync, which could lead to the PBS client apps not seeing the new data. The workaround is to include the partition key field in the object model and manually fill in this value if the data should be seen by PBS client apps.

krollins-mdb: @michael-wb, I'm not sure I follow.
> once the client app has been updated to use flexible sync
> which could lead to the PBS client apps not seeing the new data

Both can't be true, can they? The PBS clients should update to Flexible Sync.

source/sync/migrate-sync-modes.txt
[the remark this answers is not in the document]
kmorkos: It won't trigger any errors, the handlers will just always report 100% progress. So they can keep it in, it will just be unreliable

source/sync/migrate-sync-modes.txt
kmorkos: The divergence is mainly with the permission behavior in non-sync services. An `%or` expression in an app services rule that works in flexible sync will not work outside of sync and vice versa 🥴. This is a known issue that's currently being addressed, so once that's fixed we _may_ be able to migrate those scenarios automatically (can't promise that though!)

source/sync/migrate-sync-modes.txt
michael-wb: Hopefully this is clearer:

When a PBS client app has migrated to use flexible sync, the partition field will be populated with the partition value automatically by the server. The PBS client app in the migrated state relies on this partition value to subscribe to the data to download from the server.
One the client app has been updated to natively use flexible sync (i.e., no longer a PBS app), the server will no longer automatically populate the partition field. Any PBS versions of the client app still out in the wild will not see/download any objects created by the FLX version of the client app unless the partition field is manually populated when the object is created.

michael-wb: It is not guaranteed that users will update their apps (at the same time or ever) to the new FLX version.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,356 @@
...
+
+- :json-operator:`%function` operator.
+
+  :ref:`Function rules <pbs-function-rules>` are not compatible with Flexible Sync 
+  and cannot be migrated.
+
+- :json-expansion:`%%partition` expansion.
+
+  App Services Rules do not have an equivalent expansion for  
+  :json-expansion:`%%partition`, so it cannot be migrated.
```
kmorkos, line 186: we actually do migrate this automatically if it's not used in conjunction with the other restrictions listed here.
e.g. `{ %%partition: %%user.id }` becomes `{ <partitionKey>: %%user.id }`

if they need to manually update their permissions, then this is implied in the "See the list of Flexible Sync-compatible expansions for all supported expansions." note below so we can probably remove the bullet here, or change up the wording/move it elsewhere in this section

source/sync/migrate-sync-modes.txt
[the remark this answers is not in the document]
krollins-mdb: I think I'll leave this where it is, then. It's not great to have unnecessary client code, but removing it isn't exactly required.

source/sync/migrate-sync-modes.txt
krollins-mdb: Ahhh, that makes perfect sense. Thank you!

source/sync/migrate-sync-modes.txt
krollins-mdb: OK, I was wondering if that was the case. Thank you for clarifying!

source/sync/migrate-sync-modes.txt
krollins-mdb: Yeah, I think this info will be helpful for the SDK docs. I've captured this info in the Jira ticket for the SDK docs work.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+   enable Flexible Sync in the future.
+
+.. _realm-sync-migrate-commit:
+
+Commit Migration
+~~~~~~~~~~~~~~~~
+
+After your migration has completed, it is in an evaluation state. You can manually
+commit the migration to make it permanent or revert the migration to return your
+App to Partiton-Based Sync.
```
sudssm, line 230: partition**

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+You can see your migration's stage in the App Services UI. There are three stages
+to a Sync mode migration:
+
+#. **Syncing Partition-Based Sync metadata** 
+   
+   At this stage, client apps can connect to your App's backend, but local writes 
+   will not sync. These writes will be lost if you don't have a 
+   :ref:`client reset handler <handle-a-client-reset>` to recover unsynced changes.
+	 
+   Clients can also receive updates to synced collections, but they won't
```
sudssm, line 91: I'd say "clients will receive updates to synced collections, but won't be able to send uploads or receive upload acknowledgements".

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+
+- In the App Services UI, the Rules and Schema pages will be read-only.
+- Your App will experience a period of read-only Sync and then a 
+  period of downtime.
+- The new Flexible Sync configuration will set 
+  :ref:`client max offline time <client-maximum-offline-time>` to 30 days.
+- Storage in your Atlas cluster used by Device Sync metadata is temporarily doubled.
+  Partition-Based Sync metadata and Flexible Sync metadata must temporarily co-exist. 
+  Depending on the size of your synced data, this could affect billing every time you
+  migrate. The doplicated metadata is removed after the migration is rolled back
```
sudssm, line 135: duplicated*

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+   Clients will not be able to sync their local writes until the migration has
+   completed.
+
+After a migration is completed, Partition-Based Sync client apps will go through
+a client reset. Then, they can communicate with your App's Flexible Sync backend.
+You can disable and re-enable Flexible Sync and client apps will still
+communicate with your backend.
+
+If you re-enable Partitition-Based Sync you will lose the "migrated" status.
+This means you'll need to to another migration if you want to use Flexible Sync
```
sudssm, line 162: "to perform", or "to do"

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+- The injected field does not sync back to the client that created the object.
+  This means the originating client won't be able to read the injected
+  ``partitionKey`` field.
+
+The Partition-Based Sync client automatically creates sync subscriptions for 
+each table where ``partitionKey == partitionValue``. This will continue
+until you :ref:`migrate your client app <realm-sync-migrate-client>` to
+Flexible Sync.
+
+After migrating your client app to Flexible Sync, the backend will stop injecting
```
sudssm, line 316: I'd say "client app code" again here. It's also not super clear what this means I think - maybe we can say "objects created in new versions of your app will only be available to older versions if you include the `partitionKey` field" (needs some wordsmithing but something like that)

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+the partition key field. This means objects created after migrating your client
+app to Flexible Sync may not be available in clients that haven't yet been
+updated to Flexible Sync.
+
+.. _realm-sync-migrate-client:
+
+Migrate Client App to Flexible Sync
```
sudssm, line 323: again maybe "Client App Code"

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+=========================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 3
+   :class: singlecol
+
+Atlas App Services Apps can have one of two Device Sync modes: Partition-Based Sync
+and Flexible Sync. Partition-Based Sync is an older mode and you should consider
```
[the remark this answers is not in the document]
krollins-mdb: We use "is an older mode" in other places throughout the docs. I think we were hedging our bets a bit because we weren't sure what product's EOL plan for Partition-Based Sync is. Has it been communicated anywhere other than docs that PBS may be deprecated?

source/sync/migrate-sync-modes.txt
krollins-mdb: Nice catch! Thanks for pointing this out.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+
+After your app's backend has been migrated to Flexible Sync, we recommend
+that you migrate your client app to Flexible Sync. When you migrate your
+client app to Flexible Sync, the client will stop automatically creating
+sync subscriptions for each Realm object model in your app's data model.
+
+We also recommend removing all sync subscriptions and then creating new subscriptions
+for your data. This is the clearest way to control what data is synced after a
+migration. If you want to remove individual subscriptions, the automatically
+generated subscriptions use this naming format: ``flx_migrated_<objecModelName>``.
```
michael-wb, line 337: objectModelName?

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+- React Native SDK
+   :ref:`Flexible Sync - React Native SDK <react-native-flexible-sync>`
+
+- Swift SDK
+   :ref:`Manage Flexible Sync Subscriptions - Swift SDK <swift-manage-flexible-sync-subscriptions>`
+
+You can keep the automatically-generated subscriptions if you don't want to
+remove and recreate them. But you should consider that because these 
+subscriptions aren't created in your client code, they may make it harder to
+maintain and extend client code.
```
michael-wb, line 365: And they will not exist on any clients that never connected to the App after the migration to Flexible Sync was performed.

source/sync/migrate-sync-modes.txt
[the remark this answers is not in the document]
michael-wb: What about "Partition-Based Sync may be deprecated in the future and you should..."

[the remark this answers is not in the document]
michael-wb: Should you also include ref links for more information about the two modes?

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+  migration or :ref:`reverted <realm-sync-migrate-revert>` within a certain 
+  timeframe after the migration is finished.
+- Progress notifications will not work after migration because Flexible Sync does
+  not support them.  
+- Client apps that were connected to the Partition-Based Sync backend will 
+  experience a client reset after the backend is migrated to Flexible Sync. You 
+  should use a :ref:`client reset <handle-a-client-reset>` handler with recovery enabled.
+- Creating new tables in a migrated Flexible Sync backend using 
+  :ref:`Development Mode <development-mode>` from a Partition-Based Sync client
+  may lead to unexpected behavior.
```
michael-wb, line 75: Or making changes to the current schema?

Possible update:
Making changes to the current schema or creating new tables using :ref:``Development Mode <development-mode>`` in a migrated Flexible Sync backend may lead to unexpected behavior for the Partition-Based Sync clients.

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,362 @@
...
+- How long a migration takes is directly proportional to the amount of data that
+  needs to be migrated. The more data, the longer it will take to migrate.
+- Migrations can be :ref:`canceled <realm-sync-migrate-cancel>` during the 
+  migration or :ref:`reverted <realm-sync-migrate-revert>` within a certain 
+  timeframe after the migration is finished.
+- Progress notifications will not work after migration because Flexible Sync does
+  not support them.  
+- Client apps that were connected to the Partition-Based Sync backend will 
+  experience a client reset after the backend is migrated to Flexible Sync. You 
+  should use a :ref:`client reset <handle-a-client-reset>` handler with recovery enabled.
```
michael-wb, line 72: ... with recovery enabled to avoid losing any pending changes from before the client reset.

source/sync/migrate-sync-modes.txt
kmorkos: IIRC this was specifically about the PBS client not creating the subscriptions on the new tables, so the only relevant schema change here is adding a new table. I think the current wording is correct.

source/sync/migrate-sync-modes.txt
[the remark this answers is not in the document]
krollins-mdb: I don't think docs are a good place to mention that PBS may be deprecated in the future. Unless product has an EOL plan I haven't heard of (entirely possible), it's pure speculation at this point.

Good point about adding ref links. I'll add those.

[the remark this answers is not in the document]
krollins-mdb: I can follow up with Ian when he's back from vacation.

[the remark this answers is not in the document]
kmorkos: We do have a ticket to disallow enabling PBS on new sync apps which effectively makes it "deprecated", but I agree this is maybe not the right place for users to be hearing this for the first time

source/sync/migrate-sync-modes.txt
krollins-mdb: The `<placeHolder>` bit is how we indicate placeholder variable text in the docs. The idea being the tables are created based on client object model names, right? Maybe a specific example, is less confusing.

I'll take another pass on this.

kmorkos: i assume @michael-wb is calling out the typo here? it's a sneaky one
```suggestion
generated subscriptions use this naming format: ``flx_migrated_<objectModelName>``.
```

michael-wb: Sorry - your text says `<objecModelName>`, I wasn't sure if you meant to not include the `t`...

krollins-mdb: oooooh. lmao. I've rewritten to make it clearer anyway. 😆

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,376 @@
...
+
+Migration results in Partition-Based Sync clients communicating with a
+Flexible Sync backend. We recommend eventually :ref:`migrating your client app code
+<realm-sync-migrate-client>` to also use Flexible Sync instead of
+Partition-Based Sync.
+
+Requirements
+------------
+
+Currently, the Kotlin, SDK supports migrating your App's backend from
```
kmorkos:
```suggestion
Currently, the Kotlin SDK supports migrating your App's backend from
```

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,376 @@
...
+~~~~~~~~~~~~~~~
+
+Before you begin a migration, make sure you're aware of the
+:ref:`migration stages <realm-sync-migrate-stages>` and
+:ref:`effects <realm-sync-migrate-what-to-know>`.
+
+Migration generally has similar effects as :ref:`terminating <terminating-realm-sync>` 
+and re-enabling Device Sync. There are some additional effects of migration:
+
+- In the App Services UI, the Rules and Schema pages will be read-only.
```
kmorkos, line 134: not just rules + schema, _all_ config pages will be read-only (functions, auth config, triggers, etc.)

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,376 @@
...
+
+#. In the App Services UI, go to your App and select the Device Sync page.
+#. At the top of the page, select :guilabel:`[Migrate to Flexible Sync]`.
+   
+   .. image:: /images/begin-migration.png
+      :alt: Begin a migration
+   
+#. Read the information about migrating, then select 
+   :guilabel:`Next: Migration Requirements`.
+#. Ensure you meet the migration requirements, the select
```
kmorkos, line 155:
```suggestion
#. Ensure you meet the migration requirements, then select 
```

source/sync/migrate-sync-modes.txt
```diff
@@ -0,0 +1,376 @@
...
+- Swift SDK
+   :ref:`Manage Flexible Sync Subscriptions - Swift SDK <swift-manage-flexible-sync-subscriptions>`
+
+You can keep the automatically-generated subscriptions if you don't want to
+remove and recreate them. But you should consider that because these 
+subscriptions aren't created in your client code, they may make it harder to
+maintain and extend client code. 
+
+Automatically-generated subscriptions will not exist on clients that never
+connect to your App after it has been migrated to Flexible Sync.
```
kmorkos, line 377: Can't remember the context around why this note was added, but it's unclear why it's important if the client is not re-connecting? Maybe it's referring to a client app that connects as FLX after the migration without ever having connected as PBS? Probably worth clarifying if that's the case

krollins-mdb: Agreed, it's not clear. I think we could safely remove, as this should be self-evident based on all the behavior described elsewhere on the page.

PR #420, opened 2023-03-30
DOCSP-28344 App Services Dependencies Examples

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-28344

### Staged Changes

- [External Dependencies](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-28344/functions/dependencies/#upload-the-dependency-archive)
- [Realm CLI v1](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-28344/cli/realm-cli-reference-v1/index/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/cli/realm-cli-push.txt
+++ b/source/cli/realm-cli-push.txt
@@ -27,6 +27,8 @@ Syntax
 
    realm-cli push [options]
 
+.. _realm-cli-push-options:
+
 Options
 -------
 
--- a/source/cli/realm-cli-push.txt
+++ b/source/cli/realm-cli-push.txt
@@ -27,8 +27,6 @@ Syntax
 
    realm-cli push [options]
 
-.. _realm-cli-push-options:
-
 Options
 -------
 
--- a/source/cli/realm-cli-reference-v1.txt
+++ b/source/cli/realm-cli-reference-v1.txt
@@ -249,7 +249,7 @@ that doesn't exist, {+cli-bin+} can :ref:`create the application
    If specified, uploads and deploys any static assets in the
    ``/hosting/files`` directory of your App.
 
-.. option:: --include-dependencies
+.. option:: --include-node-modules
    
    | **Optional.**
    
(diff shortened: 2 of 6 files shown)


cbush (approved):

source/cli/realm-cli-reference-v1.txt
```diff
@@ -249,7 +249,7 @@ that doesn't exist, {+cli-bin+} can :ref:`create the application
    If specified, uploads and deploys any static assets in the
    ``/hosting/files`` directory of your App.
 
-.. option:: --include-dependencies
+.. option:: --include-node-modules
```
dacharyc, line 252: This file is for realm-cli version 1. This is an older version of realm-cli that I do not believe is getting updates anymore. Before we change this flag in the v1 reference, we should try to install a v1 realm-cli and see if this flag has actually been updated there.

Separately, we may want to remove the v1 reference as it has now been quite a while since realm-cli v2 was released and I'm not sure how much this is needed anymore. It may cause more confusion than help at this point since there is now a much more comprehensive v2 reference.

dacharyc (commented):
Can we verify that the realm-cli v1 reference change is actually correct before we merge this?

source/cli/realm-cli-push.txt
```diff
@@ -27,6 +27,8 @@ Syntax
 
    realm-cli push [options]
 
+.. _realm-cli-push-options:
```
dacharyc, line 30: This is a good instinct! Unfortunately, the `realm-cli-push.txt` page is one of the documents that is automatically generated whenever we regenerate the realm-cli reference documentation. So if you manually add this target, we'll lose it again the next time we regenerate these docs. (Everything in the CLI directory aside from the realm-cli-reference-v1 is autogenerated. If we manually update these pages, we lose any changes the next time we regenerate these docs.)

I don't think we really need this link at all, so you can go ahead and remove this from here and remove the `:ref:` in ln 299/300 in the dependencies.txt file.

dacharyc (commented):
Looking good! Just a couple more tiny tweaks.

dacharyc (approved):
Awesome - looks great! Nice work 👍

PR #439, opened 2023-04-19
(DOCSP-29274) [Admin API] Private Endpoints

nlarew:
### Jira

- (DOCSP-29274) [Admin API] Private Endpoints

### Staged Changes

- [Admin API > Security > VPC Private Endpoints](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/privatelink/admin/api/v3/#tag/security)

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -7806,9 +7806,10 @@ components:
           type: string
         dns:
           type: string
+          example: "*.aws.realm.mongodb.com"
         service_name:
           type: string
-
+          example: com.amazonaws.vpce.us-east-1.vpce-svc-0f12ab34cd56ef789
 
   securitySchemes:
     tokenAuth:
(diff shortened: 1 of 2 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -2360,6 +2360,112 @@ paths:
...
+    get:
+      operationId: adminGetPrivateEndpoint
+      tags:
+        - security
+      summary: Get a VPC Private Endpoint
+      description: |
+        Get a specific private endpoint configuration.
+      responses:
+        200:
+          description: Found
```
dacharyc, line 2416: The description says "Get a specific private endpoint configuration" but the response is just 200/Found - do we have an example we could add, or does it not return anything aside from the code?

dacharyc (approved):
Overall, LGTM - just one question about possibly providing an additional response example.

PR #441, opened 2023-04-20
(DOCSP-24695): Create App Services tutorial for Triggers and Atlas Search

davidhou17:
### Jira

- https://jira.mongodb.org/browse/DOCSP-24695

### Staged Changes

- [Build an Application using Triggers and Atlas Search](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-24695/tutorial/triggers-fts/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/tutorial/triggers-fts.txt
+++ b/source/tutorial/triggers-fts.txt
@@ -1,7 +1,7 @@
 .. _triggers-fts-tutorial:
 
 ====================================================
-Build an Application using Triggers and Atlas Search
+Build an Application Using Triggers and Atlas Search
 ====================================================
 
 .. contents:: On this page
@@ -58,7 +58,7 @@ Before you begin:
 
 .. include:: /includes/tutorial-template-prerequisite.rst 
 
-- To run Atlas Search queries, ensure that your Atlas account runs MongoDB 
+- To run Atlas Search queries, ensure that your Atlas cluster runs MongoDB 
   version 4.2 or higher.
 
 Create a New Template App
@@ -130,14 +130,14 @@ Store Alert Terms in Atlas
 --------------------------
 
 Now that you've created the frontend component for the alert box, 
-configure the application's backend operations so that you can store 
-and keep track of your alert terms in Atlas.
+configure the application's backend to store and keep track of your 
+alert terms in Atlas.
 
 .. procedure::
 
    .. step:: Create a New Collection
 
-      Create a collection that will store the terms that users enter in the app.
+      Create a collection to store the terms that users enter in the app.
       In the Atlas UI: 
 
       A. Click the :guilabel:`Data Services` tab.
@@ -158,7 +158,7 @@ and keep track of your alert terms in Atlas.
 
       A. Click the :guilabel:`App Services` tab.
       
-      #. Click the tile for your template app.
+      #. Click the tile for your app.
 
       #. In the left navigation under :guilabel:`Data Access`, click :guilabel:`Rules`.
 
@@ -171,7 +171,7 @@ and keep track of your alert terms in Atlas.
    .. step:: Write to the Collection
 
       Now that you've configured write access to the ``todo.alerts`` collection, 
-      create a function to write to the collection. The following TypeScript function 
+      create a helper function to write to the collection. The following TypeScript function 
       takes a string input and uses the :ref:`React Native SDK <react-native-mongodb-data-access>` 
       to connect to Atlas and insert the specified alert term as a document to your collection.  
 
@@ -238,7 +238,7 @@ function, you define:
       To open the database trigger configuration page in the App Services UI:
 
       A. Click the :guilabel:`App Services` tab and select the tile for your 
-         template app.
+         app.
 
       #. In the left navigation menu, click :guilabel:`Triggers`.
       
--- a/source/tutorial/triggers-fts.txt
+++ b/source/tutorial/triggers-fts.txt
@@ -104,7 +104,7 @@ might appear in your important or time-sensitive tasks.
 
       .. code-block:: javascript
       
-         import {AlertBox} from './AlertBox'
+         import {AlertBox} from './AlertBox';
 
       Then, add the following lines to your code to render a button that 
       displays the alert box when clicked:
@@ -119,7 +119,7 @@ might appear in your important or time-sensitive tasks.
          :caption: src/ItemListView.tsx
          :emphasize-lines: 19-41
 
-      To change the color of the button, add the following lines to your code:
+      To change the size and color of the button, add the following lines to your code:
 
       .. literalinclude:: /tutorial/backend-examples/ItemListView.snippet.alertButtonStyle.tsx
          :language: javascript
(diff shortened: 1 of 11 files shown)


davidhou17:
@cbush Not sure if you were in the middle of your review, but I incorporated the feedback you had left and left a follow up comment when you're ready for another look! Thanks!

source/tutorial/triggers-fts.txt
```diff
@@ -0,0 +1,345 @@
...
+      
+         import {AlertBox} from './AlertBox';
+
+      Then, add the following lines to your code to render a button that 
+      displays the alert box when clicked:
+
+      .. literalinclude:: /tutorial/backend-examples/ItemListView.snippet.showAlertBox.tsx
+         :language: javascript
+         :caption: src/ItemListView.tsx
+         :emphasize-lines: 2
```
nlarew, line 115: A newbie react dev might be a bit confused as to where these two lines go. Can we expand the snippet to show a bit more context or at least give some helpful text? (e.g. you define them on the first and second line of the function block)

source/tutorial/triggers-fts.txt
```diff
@@ -0,0 +1,345 @@
...
+
+      #. In the left navigation under :guilabel:`Data Access`, click :guilabel:`Rules`.
+
+      #. Under the ``todo`` database, click the ``alerts`` collection.
+      
+      #. In the right dialog box, select the :guilabel:`readAndWriteAll` preset role.
+      
+      #. Click :guilabel:`Add preset role` to confirm your selection.
+      
+   .. step:: Write to the Collection
```
nlarew, line 171: It wasn't immediately clear that this jumps back to the RN app code rather than still being in the App Services UI. Can we make that clearer?

source/tutorial/triggers-fts.txt
```diff
@@ -0,0 +1,345 @@
...
+      
+      #. Click the tile for your app.
+
+      #. In the left navigation under :guilabel:`Data Access`, click :guilabel:`Rules`.
+
+      #. Under the ``todo`` database, click the ``alerts`` collection.
+      
+      #. In the right dialog box, select the :guilabel:`readAndWriteAll` preset role.
+      
+      #. Click :guilabel:`Add preset role` to confirm your selection.
```
nlarew, line 169: Might want to add a blurb that if you're using deployment drafts in the UI then you need to deploy before the changes are live. (I'm not actually sure if these are turned on by default for the RN template backend)

source/tutorial/triggers-fts.txt
```diff
@@ -0,0 +1,345 @@
...
+         - Enable :guilabel:`Full Document` to include each new report document
+           in the change event passed to the trigger function.
+
+   .. step:: Define an Atlas function
+
+      To define the trigger's function, navigate to the :guilabel:`Function` 
+      section of the configuration page and select :guilabel:`+ New Function` 
+      from the drop-down menu. Name the function ``triggers/sendAlerts``.
+      
+      The following JavaScript function returns a message in your application
```
nlarew, line 270: I read "the following JS function" but then had to actually scroll the page down past some text to see the code. The code block also has an intro sentence that uses "the following" too. Can we rework this section so that "the following" only comes immediately before the code block?

One approach - we can move the code block up and then have all the explanation below it. This would also help to highlight the currently orphaned "click save and deploy" instruction at the end of the step.

e.g.

1. Name the function `triggers/sendAlerts`
2. Copy the following code into the function body
3. Here's what it does....
4. Click save and deploy

source/tutorial/triggers-fts.txt
```diff
@@ -0,0 +1,345 @@
...
+            "The following task has been added to a to-do list: Complete important tutorial.
+            You've been alerted because it contains the term, important.",
+            "Related incomplete tasks: [
+            {"summary": "Important: Create template app"}, 
+            {"summary": "Add important tasks"},
+            {"summary": "Make sure to read the documentation. This is important."}]"
+         ]
+                  
+What's Next?
+------------
```
nlarew, line 333: One other thing we might add - "extend the alert function to call an external service that can send you a text or an email instead of simply logging"

nlarew (commented):
Cool tutorial! Looking solid, just a few suggestions

source/tutorial/triggers-fts.txt
davidhou17: Ah, I had development mode on so drafts were disabled - good catch!

davidhou17:
@nlarew Thanks for the fantastic feedback! Back to you for another look when you get a chance!

nlarew (approved):
LGTM

eshwer:
This is super cool, are we able to double link it from our tutorials as well?

source/tutorial/triggers-fts.txt
```diff
@@ -0,0 +1,365 @@
...
+applications.
+
+In this tutorial, you start with a pre-built TypeScript mobile application 
+that includes a working React Native application (frontend) and its corresponding 
+App Services App configuration files (backend). This template app is a basic 
+to-do list application that lets users do various things to manage their tasks.
+To learn more about the template app, see :ref:`react-native-tutorial`.
+
+After you've got the template app running, you will add a new feature that alerts you 
+when a user creates a task that contains a specific word or phrase. The feature
```
awjian, line 27: Confirming that this template app allows multiple users to create tasks, correct? Asking because I'd like to make sure that the _use case_ is clear here, ie. User A gets notified if User B adds a task with the term "urgent" in it. It may also help to give an example of what type of alert terms users might find to be useful at the beginning of the tutorial instead of only at the end.

awjian (commented):
few minor comments, great work!

source/tutorial/triggers-fts.txt
davidhou17: Yes, that's correct! This is a great piece of feedback - will include an example in the intro now

davidhou17:
@awjian Back to you when you get a chance to take another look!

source/tutorial/triggers-fts.txt
```diff
@@ -0,0 +1,365 @@
...
+      #. By default, your application enables deployment drafts. To 
+         :ref:`manually deploy <deploy-ui>` your changes, click 
+         :guilabel:`Review Draft & Deploy` and then :guilabel:`Deploy`.
+         
+   .. step:: Write to the Collection
+
+      Once you've configured write access to the ``todo.alerts`` collection, return
+      to your application code.
+
+      In ``src/ItemListView.tsx``, add the following lines to the top of tbe function block
```
laurazhukas, line 193: small nitpick - typo here t**b**e instead of the

laurazhukas (commented):
Tutorial looks great! Thanks for all the hard work y'all have put in - I dropped one small comment, otherwise lgtm 👏

source/tutorial/triggers-fts.txt
davidhou17: Good catch - thanks!!

laurazhukas (approved):
lgtm!

awjian (approved):
LGTM! Thanks for the great work on this one

awjian (approved):
LGTM! Thanks for the great work on this one

cbush (approved):

PR #450, opened 2023-04-28
(DOCSP-29512) Use an OpenAPI bundler to include examples in Data API spec

nlarew:
## Pull Request Info

- Switches to a new format where we have split OpenAPI definitions that we bundle before publishing
- Includes tested Bluehawk Data API cURL snippets directly in the OpenAPI spec
- Adds a GitHub action that ensures we bundled the spec for production & that the spec is valid
- Adds a README that explains what's going on

### Jira

(DOCSP-29512) Use an OpenAPI bundler to include examples in Data API spec

### Staged Changes

- [Data API > OpenAPI](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/openapi-data-api/data-api/openapi/)


openapi-data-api-v1/preview-bundled.sh
```diff
@@ -0,0 +1,5 @@
+cd "$(dirname "$0")"
+
+BUNDLED_OPENAPI_FILE="../openapi-data-api-v1.bundled.yaml"
+
+npx @redocly/cli@latest preview-docs "$BUNDLED_OPENAPI_FILE"
```
mongodben, line 5: rather than have these 2 bash scripts, consider making a node.js project for this.

and then both these scripts can be in a package.json file.

this has the added bonus of being able to use [nodemon](https://www.npmjs.com/package/nodemon) (or smthn similar) to watch these files for changes and auto-update the preview.

mongodben (approved):
lgtm but think it could be enhanced with https://github.com/mongodb/docs-app-services/pull/450/files#r1181677905

openapi-data-api-v1/preview-bundled.sh
nlarew: Sure, switched to node. Don't need nodemon though, the CLI preview already watches your files.

PR #459, opened 2023-05-04
(DOCSP-28050): Clarify match expression

cbush:
- Note that most of what is being asked for in the ticket was already done elsewhere.
- I don't think we should recommend fullDocumentBeforeChange because it requires special cluster configuration.

### Jira

- https://jira.mongodb.org/browse/DOCSP-28050

### Staged Changes

- [Database Triggers](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/match-clarity/triggers/database-triggers/#use-match-expressions-to-limit-trigger-invocations)

### DIFF
--- a/source/triggers/database-triggers.txt
+++ b/source/triggers/database-triggers.txt
@@ -635,14 +635,12 @@ Expression` field. App Services evaluates the match expression against the
 change event document and invokes the Trigger only if the expression evaluates
 to true for the given change event.
 
-The "match expression" is basically a JSON document that specifies the query
-conditions using the :manual:`MongoDB read query syntax
-</reference/operator/query/>`.
+The match expression is a JSON document that specifies the query conditions
+using the :manual:`MongoDB read query syntax </reference/operator/query/>`.
 
 We recommend only using match expressions when the volume of Trigger events
 measurably becomes a performance issue. Until then, receive all events and
-decide what to do (if anything) with any given event in the Trigger function
-code.
+handle them individually in the Trigger function code.
 
 The exact shape of the change event document depends on the event that caused
 the trigger to fire. For details, see the reference for each event type:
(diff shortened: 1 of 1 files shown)


source/triggers/database-triggers.txt
```diff
@@ -635,9 +635,14 @@ Expression` field. App Services evaluates the match expression against the
 change event document and invokes the Trigger only if the expression evaluates
 to true for the given change event.
 
+The "match expression" is basically a JSON document that specifies the query
```
MongoCaleb, line 638: remove "basically"

source/triggers/database-triggers.txt
```diff
@@ -635,9 +635,14 @@ Expression` field. App Services evaluates the match expression against the
...
+conditions using the :manual:`MongoDB read query syntax
+</reference/operator/query/>`.
+
 We recommend only using match expressions when the volume of Trigger events
```
MongoCaleb, line 642: We have "match expression" in quotes above and not here. I think it's harder to scan here without the quotes.

source/triggers/database-triggers.txt
```diff
@@ -635,9 +635,14 @@ Expression` field. App Services evaluates the match expression against the
...
 measurably becomes a performance issue. Until then, receive all events and
-implement the logic to ignore certain events in the Trigger Function.
+decide what to do (if anything) with any given event in the Trigger function
```
MongoCaleb, line 644: Not sure what "with any given event" means. I think it's referring to the list that follows, but it's not clear. Could we try:
> Until then, receive all events and handle them individually in the Trigger function.
?

MongoCaleb (commented):
A couple of wording suggestions.

source/triggers/database-triggers.txt
cbush: I will remove quotation marks, this is the only place I used them (for my own reasons)

MongoCaleb (approved):
Lgtm

PR #470, opened 2023-05-11
(DOCSP-29722) Improved unordered trigger throughput

nlarew:
### Jira

- (DOCSP-29722) Improved unordered trigger throughput

### Staged Changes

- [Triggers > Limitations > Event Processing Throughput](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/triggers-throughput/triggers/#event-processing-throughput)

### DIFF
--- a/source/triggers.txt
+++ b/source/triggers.txt
@@ -71,23 +71,25 @@ constraints as all Atlas Functions.
 
 :ref:`Learn more about Atlas Function constraints.<function-constraints>`
 
-Maximum Executions Per Second
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Event Processing Throughput
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-App Services limits the execution of Trigger
-Functions. A Trigger can execute up to 2500 times per second. When a
-Trigger calls a Function, this call goes into a queue. When capacity
-becomes available, App Services executes the Function call.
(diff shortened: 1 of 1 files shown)


source/triggers.txt
```diff
@@ -71,23 +71,24 @@ constraints as all Atlas Functions.
...
-Database Triggers with *event ordering* work differently. Each ordered
-Trigger processes events in sequence. App Services waits for each event
-execution in the sequence to complete. Then, it handles the next event. Only
-one sequence for a given ordered Trigger executes at a time.
+- Optimize the Trigger function's run time. For example, you might
+  reduce the number of network calls that you make.
+
+- Reduce the size of each event object with the Trigger's projection
+  filter. For the best performance, we recommend limiting the size of
+  each change event to 2KB or less.
```
cbush, line 91: Hmm, I wonder if another suggestion would be to use match expressions like we recommend on the [Database Triggers](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/triggers-throughput/triggers/database-triggers/#use-match-expressions-to-limit-trigger-invocations) page.

cbush (approved):

source/triggers.txt
nlarew: I think it's reasonable if you're already doing matching logic in the function. Then it's throughput improvement by not taking up a slot in the 10k batch.

PR #503, opened 2023-06-15
 Backend Sync mode migration: Realm.js support 

krollins-mdb:
## Pull Request Info

The React Native and Node.js SDKs now support backend Sync mode migration.

- [Migrate Device Sync Modes](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/sync-migration-js/sync/migrate-sync-modes/#requirements)

### Review Guidelines

### DIFF
--- a/source/sync/migrate-sync-modes.txt
+++ b/source/sync/migrate-sync-modes.txt
@@ -27,9 +27,8 @@ Partition-Based Sync.
 Requirements
 ------------
 
-Currently, the Swift, Kotlin, and Java client SDKs support migrating your App's
-backend from Partition-Based Sync to Flexible Sync. The .NET, Node.js, and
-React Native SDKs do not support backend Sync mode migration yet.
+The .NET SDK does not support backend Sync mode migration yet. This feature does
+not apply to clients that use the C++, Flutter, and Web SDKs.
 
 App Services App:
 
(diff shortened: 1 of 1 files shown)


source/sync/migrate-sync-modes.txt
```diff
@@ -27,9 +27,8 @@ Partition-Based Sync.
 Requirements
 ------------
 
-Currently, the Swift, Kotlin, and Java client SDKs support migrating your App's
-backend from Partition-Based Sync to Flexible Sync. The .NET, Node.js, and
-React Native SDKs do not support backend Sync mode migration yet.
+The .NET SDK does not support backend Sync mode migration yet. This feature does
+not apply to clients that use the C++, Flutter, or Web SDKs.
```
dacharyc, line 31: Should we mention _why_ this feature isn't available in those SDKs? (i.e. they have never supported Partition-Based Sync?)

I just wonder if people will see them called out here and wonder if there's a plan to add them or otherwise wonder why they're called out.

dacharyc (approved):
LGTM! I have a Q about calling out the SDKs where this feature is not available, but I don't feel strongly about it so I'm happy to leave it up to you whether to tweak it.

source/sync/migrate-sync-modes.txt
krollins-mdb: I initially added that detail in, but felt like it was too much info for what's effectively an admonition in the "Requirements" section. Looking at it now, it may be best to not mention them at all. 🤔

krollins-mdb:
@dacharyc, after thinking about it some more, I removed the reference to the non-applicable SDKs altogether. Does it feel better to not address these SDKs at all?

dacharyc:
> @dacharyc, after thinking about it some more, I removed the reference to the non-applicable SDKs altogether. Does it feel better to not address these SDKs at all?

I'm really torn. It does seem conspicuous that they're missing. If you're a newcomer to Realm and don't know/understand that these SDKs have never supported Partition-Based Sync, it may make you wonder why they're missing.

But if you're a newcomer, you don't need this feature _anyway_ because you're probably already using Flexible Sync. 🤔 

Let's just leave it out for now, and then if somebody comments later, we can always add something about it.

krollins-mdb:
Sounds good. Thanks for thinking it through with me!

PR #515, opened 2023-07-05
Update text and image to match UI changes for Last Heartbeat

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-30418

### Staged Changes

https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-30418/triggers/database-triggers/#manually-resume-a-suspended-trigger


source/triggers/database-triggers.txt
```diff
@@ -466,18 +466,19 @@ importing an application directory with {+cli-bin+}.
...
+is not sending any events, the server sends a heartbeat to ensure the trigger's 
+resume token stays fresh. Whichever event is most recent is shown as the 
+:guilabel:`Latest Heartbeat` in the App Services UI. Database Triggers also 
+keep track of the :guilabel:`Last Cluster Time Processed`, which is the last time 
+the change stream backing that Trigger emitted an event. You can access these 
+timestamps in the list of Triggers that displays when you first navigate to the
 :guilabel:`Triggers` section from the sidebar.
 
-The difference between the latest execution and the last cluster time
+The difference between the latest heartbeat and the last cluster time
```
tkaye407, line 478: I believe this second statement is no longer true. The heartbeat can be now and the last cluster time can be a day ago and that just means it has been 1 day since an event, right? 

cc @Calvinnix and @dpish for clarity

MongoCaleb: Thanks. Because we define it (a bit incorrectly) in the paragraph above, I'm gong to clean this whole section up just a bit.

MongoCaleb: @dpish 
You say "The heartbeat can be now(), it indicates the last time the trigger attempted to process any available events."
I thought that if there are no events, we send a heartbeat, and this value might just represent that, not just the last time an event was handled. 
As for the timing on Last Cluster Time, are you saying it can't be more than 1 day, or are you just using that as an example?

MongoCaleb: In other words, is this correct? 
> **Latest Heartbeat**,

Atlas App Services keeps track of the last time a trigger was run. If the trigger 
is not sending any events, the server sends a heartbeat to ensure the trigger's 
resume token stays fresh. Whichever event is most recent is shown as the 
:guilabel:`Latest Heartbeat`.

**Last Cluster Time Processed**

This is the time when the trigger last processed an event. It will be older than 
the :guilabel:`Latest Heartbeat` if there have been no events since the most 
recent heartbeat.

dpish: Last Heartbeat looks great.

For Last Cluster Time Processed, what you had earlier is correct:
> Database Triggers also keep track of the :guilabel:`Last Cluster Time Processed`, which is the last time 
> the change stream backing that Trigger emitted an event. 

So it could be something like:
> Database Triggers keep track of the :guilabel:`Last Cluster Time Processed`, which is the last time 
> the change stream backing that Trigger emitted an event. It will be older than
> the :guilabel:Latest Heartbeat if there have been no events since the most
> recent heartbeat.

source/triggers/database-triggers.txt
```diff
@@ -463,118 +553,29 @@ importing an application directory with {+cli-bin+}.
...
-of the **last cluster time processed**: the last time the change stream
-backing that Trigger emitted an event. You can access these timestamps
-in the list of Triggers that displays when you first navigate to the
-:guilabel:`Triggers` section from the sidebar.
-
-The difference between the latest execution and the last cluster time
-processed represents the latency between when an event occurs in your
-database, and when the Trigger finishes responding to that event. If
-latency is consistently large, you can reduce it with the following
-mitigations:
```
dpish, line 480: I'm okay removing this since we dont display latency but the steps to address that are still valid, ie this bit:

> If latency is consistently large, you can reduce it with the following mitigations:
> 
> - Optimize the Trigger's linked Function to speed up execution.
> 
> - If your Trigger uses Event Ordering, consider switching ordering off
>   to reduce latency.
> 
> - Add resources by upgrading your MongoDB Atlas cluster.

dpish (approved):

tkaye407 (approved):

source/triggers/database-triggers.txt
MongoCaleb: Agreed, but unless we discuss how to know if there is latency, I think it's weird to say "to fix latency...". Any thoughts on that?

MongoCaleb: I think all of these are already covered in more detail in the "Performance Optimization" section.

source/triggers/database-triggers.txt
```diff
@@ -290,6 +289,97 @@ Database change event objects have the following general form:
...
+     shipDate: ISODate("2018-06-27T08:20:23.311Z"),
+     orderContents: [
+       { qty: 1, name: "Earl Grey Tea Bags - 100ct", price: NumberDecimal("10.99") }
+     ],
+     shippingLocation: [
+       { location: "Memphis", time: ISODate("2018-06-27T18:22:33.243Z") },
+     ]
+   }
+
+To automate this process, the store creates a database Trigger that
```
lindseymoore, line 314:
```suggestion
To automate this process, the store creates a Database Trigger that
```
Should database be capitalized too?

lindseymoore (approved):
LGTM!

PR #535, opened 2023-08-01
(DOCSP-31628) Add documentation for development mode breaking changes

krollins-mdb:
## Pull Request Info

This shouldn't be merged until the next BAAS release.

### Jira

- https://jira.mongodb.org/browse/DOCSP-31628

### Staged Changes

- [Sync Settings](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31628/sync/configure/sync-settings/#development-mode)

### Review Guidelines

### DIFF
--- a/source/includes/dev-mode-breaking-changes.rst
+++ b/source/includes/dev-mode-breaking-changes.rst
@@ -4,8 +4,8 @@ Development Mode is not suitable for production use. If you use Development
 Mode, make sure to disable it before moving your app to production.
 
 App Services Apps in Development Mode that were created after August 9th, 2023
-can make :ref:`breaking changes <breaking-change-quick-reference>` to synced
-object schemas from their client code.
+can make :ref:`breaking changes <breaking-change-quick-reference>` from their
+client code to synced object schemas.
 
 Refer to :ref:`Development Mode <dev-mode-breaking-changes>` for details about
 making breaking changes in Development Mode.
--- a/source/includes/dev-mode-breaking-changes.rst
+++ b/source/includes/dev-mode-breaking-changes.rst
@@ -1,0 +1,11 @@
+*Applies to App Services Apps created after August 9th, 2023.* 
+
+Development Mode is not suitable for production use. If you use Development
+Mode, make sure to disable it before moving your app to production.
+
+App Services Apps in Development Mode that were created after August 9th, 2023
+can make :ref:`breaking changes <breaking-change-quick-reference>` to synced
+object schemas from their client code.
+
+Refer to :ref:`Development Mode <dev-mode-breaking-changes>` for details about
+making breaking changes in Development Mode.
--- a/source/includes/dev-mode-breaking-changes.rst
+++ b/source/includes/dev-mode-breaking-changes.rst
@@ -1,11 +1,11 @@
 *Applies to App Services Apps created after August 9th, 2023.* 
 
-Development Mode is not suitable for production use. If you use Development
-Mode, make sure to disable it before moving your app to production.
-
 App Services Apps in Development Mode that were created after August 9th, 2023
 can make :ref:`breaking changes <breaking-change-quick-reference>` from their
 client code to synced object schemas.
 
 Refer to :ref:`Development Mode <dev-mode-breaking-changes>` for details about
 making breaking changes in Development Mode.
+
+Development Mode is not suitable for production use. If you use Development
+Mode, make sure to disable it before moving your app to production.
--- a/source/sync/configure/sync-settings.txt
+++ b/source/sync/configure/sync-settings.txt
@@ -78,8 +78,8 @@ Breaking Changes
 *Applies to App Services Apps created after August 9th, 2023.* 
 
 App Services Apps in Development Mode that were created after August 9th, 2023
-can make :ref:`breaking changes <breaking-change-quick-reference>` from their
-client code to synced object schemas.
+can make :ref:`breaking changes <breaking-change-quick-reference>` to synced
+object schemas from client code.
 
 .. note:: Apps created before August 9th, 2023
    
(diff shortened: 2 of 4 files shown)


dacharyc (changes requested):
I think we're missing a few places where we would want to surface this information. The "Configure and Update Your Data Model" section has a couple of pages where we talk extensively about breaking changes:

- [Update Your Data Model](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31628/sync/data-model/update-schema/)
- [Make Breaking Schema Changes](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31628/sync/data-model/migrate-schema-partner-collection/)

We should probably surface this info in both of these places. And maybe we should make a distinction about making breaking changes while the app is in production vs. development, since we tell people not to use Development Mode in Production apps - so this new feature really only applies to apps that are in development.

source/includes/dev-mode-breaking-changes.rst
```diff
@@ -0,0 +1,11 @@
+*Applies to App Services Apps created after August 9th, 2023.* 
+
+Development Mode is not suitable for production use. If you use Development
+Mode, make sure to disable it before moving your app to production.
+
+App Services Apps in Development Mode that were created after August 9th, 2023
+can make :ref:`breaking changes <breaking-change-quick-reference>` from their
+client code to synced object schemas.
```
dacharyc, line 8: I can't seem to find the grammar rules for why I think it should be this way, but I have in my head the "to" should come before the "from" in this sentence. This is definitely a nitpick, though, so YMMV.

```suggestion
can make :ref:`breaking changes <breaking-change-quick-reference>` to synced
object schemas from client code.
```

source/sync/configure/sync-settings.txt
```diff
@@ -61,15 +61,64 @@ update the backend schema by changing the client model.
...
+   object schema in the App Services UI. For details, see
+   :ref:`<synced-schema-overview>`
+
+To make a breaking change from client code:
+
+#. Change your local object model and let it sync to your App.
+#. Delete your local realm and data.
+#. Open a new realm with your updated object model.
+
+To delete a Realm file use the Realm SDK-specific methods:
```
dacharyc, line 96: If the requirement is to delete the realm, there are multiple ways we might do that. If an app is in development, presumably we can just delete the App from the simulator. Also, there are some considerations about _when_ you can use the `.delete` method in Swift. (And this delete method isn't correct for Swift - it's `Realm.deleteFiles(for: config)`)

Instead of manually supplying untested SDK methods here without helpful context/API reference links, maybe it makes sense to link out to the SDK Delete pages? i.e. https://www.mongodb.com/docs/realm/sdk/swift/realm-files/delete-a-realm/

source/includes/dev-mode-breaking-changes.rst
```diff
@@ -0,0 +1,11 @@
+*Applies to App Services Apps created after August 9th, 2023.* 
+
+Development Mode is not suitable for production use. If you use Development
```
dacharyc, line 3: This note read strangely to me on the relevant pages with the "Dev Mode not suitable for production use" sentence first. Maybe we want to move this to the bottom of the include?

dacharyc (commented):
Unless there's a good reason to list the SDK delete methods here, I think we should probably just link out to the SDK pages for Deleting a Realm. I don't love the idea of using untested SDK code in the App Services docs, and we may be missing context that is important around these methods. Curious what you think or if there's something I'm missing about why we should list them here.

source/includes/dev-mode-breaking-changes.rst
krollins-mdb: 🤣 I went back and forth on this sentence about a dozen times. I wanted to try to emphasize that the breaking changes are happening in the client, which is what the new thing is all about. If that seems too awkward, I'm happy to change it.

source/sync/configure/sync-settings.txt
krollins-mdb: Yeah... not all of the SDKs document how to delete a realm. Like the JS SDKs. *hides*

I admittedly opted for the lazier approach with doing the table of untested code snippets. Instead, it probably makes sense to link to SDK docs where it's documented and link to API reference docs where it's not. I'll update!

source/includes/dev-mode-breaking-changes.rst
krollins-mdb: I wasn't totally satisfied leading with that sentence anyway. I definitely want to make the point, but it almost seems like we're immediately telling people not to use the thing. lol

source/sync/configure/sync-settings.txt
dacharyc: I guess I know some pages that should be added to the "gaps" epic! 😂 

This seems like a reasonable compromise.

dacharyc (approved):
LGTM!

source/sync/configure/sync-settings.txt
```diff
@@ -61,15 +61,48 @@ update the backend schema by changing the client model.
...
+   
+   For Apps created before August 9th, 2023, you need to update your
+   object schema in the App Services UI. For details, see
+   :ref:`<synced-schema-overview>`
+
+To make a breaking change from client code:
+
+#. Change your local object model and let it sync to your App.
+#. Delete your local realm and data.
+#. Open a new realm with your updated object model.
```
yaseenmongo, line 94: "and let it sync to your app" is a little misleading imo. Whenever they make a change and try to run their mobile app again, it will fail to open with an error message.

The flow is to change their schema, then delete the realm before opening it with the new schema changes.

yaseenmongo (commented):
Apologies for the late review was out sick most of last week. Had one minor comment about the phrasing of things around making the change. But otherwise, everything else looks great. Super straightforward and clear.

source/sync/configure/sync-settings.txt
krollins-mdb: Ooh, great point. Thank you! I removed the sync bit in my latest (had to revert because the feature didn't go live).

krollins-mdb: Latest PR is here: https://github.com/mongodb/docs-app-services/pull/548

PR #537, opened 2023-08-02
DOCSP-31190 Zero value for missing required property

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-31190

### Staged Changes

- [Note in Required Property section](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31190/sync/data-model/update-schema/#add-a-required-property)
- [Addition to Default Row in Property Types table](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31190/sync/data-model/data-model-map/#property-types)

### Reminder Checklist

### Review Guidelines


source/sync/data-model/update-schema.txt
```diff
@@ -178,6 +178,12 @@ required property to both the client model and the Atlas schema.
 However, you should consider making the property optional to avoid the need to 
 modify existing Atlas documents. 
 
+.. note:: Objects missing a required property will default to a zero value
```
dacharyc, line 181: Nit: our style guide calls for us to avoid future tense.

```suggestion
.. note:: Objects missing a required property default to a zero value
```

source/sync/data-model/update-schema.txt
```diff
@@ -178,6 +178,12 @@ required property to both the client model and the Atlas schema.
...
+
+   Clients are allowed to create and Sync objects missing a required property. 
+   The server will populate the missing required value field with a zero or blank value
```
dacharyc, line 184: Nit: tense.
```suggestion
   The server populates the missing required value field with a zero or blank value
```

dacharyc (commented):
Two general comments:

- This page is about changing a data model. I think this behavior is general to _any_ required property - not just adding one when you update the data model - so it might be good to also include this note on the "[Data Model Mapping](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31190/sync/data-model/data-model-map/#property-types)" page.
- The "Property Types" table on the Data Model Mapping page has a "Default" parameter that explicitly says "When you attempt to create an object that is missing a value for a required field, it fails validation and does not persist to the realm." This note seems to contradict that. It would be good to confirm which behavior is expected, and update the verbiage on the page if needed.

source/sync/data-model/update-schema.txt
```diff
@@ -178,6 +178,13 @@ required property to both the client model and the Atlas schema.
 However, you should consider making the property optional to avoid the need to 
 modify existing Atlas documents. 
 
+.. note:: Required Properties Missing in a Schema Subset Will Default to Zero 
+
+   Clients can open the Realm with a subset of a schema that doesn't 
+   include a required property. In this case, the server populates the 
+   missing required value field with a zero or blank value
+   (like 0, "", or 0.0), depending on the property type.
+
```
dacharyc, line 187: I can't comment directly, but the section below this - "Modify Existing Atlas Documents" - describes how you have to update existing documents with a new required property or they don't sync with the client. 

Is that still true with this new required property behavior? i.e. does the server populate missing required value fields only for _new_ documents, or does it also retroactively apply the missing default for documents that are already in a collection?

If it's true only for _new_ documents, we should probably add that in the note. Maybe something like "When you sync objects from the client that are missing a required property, the server populates the missing required value field with..."

And then maybe add clarification in the section below, too, that this behavior doesn't apply to existing documents in a collection, and they still have to populate those with default values or they don't sync.

OR if they no longer have to add default values to existing documents to get them to sync, we should probably delete the section below?

dacharyc (commented):
One more question about something that we should probably clarify before we merge.

source/sync/data-model/update-schema.txt
lindseymoore: It's only for new documents! I think specifying in the note will suffice. Wouldn't want to convolute the next section for such a small case. Let me know what you think of the update!

dacharyc: Perfect, that makes sense. Thank you for clarifying!

dacharyc (approved):
LGTM!

source/sync/data-model/update-schema.txt
```diff
@@ -178,6 +178,13 @@ required property to both the client model and the Atlas schema.
 However, you should consider making the property optional to avoid the need to 
 modify existing Atlas documents. 
 
+.. note:: Required Properties Missing in a Schema Subset Will Default to Zero for New Documents
```
benredmond, line 181: Sorry I think I misunderstood your question on slack about this - we actually do this every time for data coming from Realm. this section LGTM if we just remove the references to "new"

benredmond (approved):
LGTM aside from a clarification on the "new" documents coming from Realm. Thanks for picking this up! I'm happy that we're making this behavior more explicit to the user :)

PR #548, opened 2023-08-14
DOCSP-31628: Add documentation for development mode breaking changes

krollins-mdb:
## Pull Request Info

This feature was pulled from the latest BAAS release. Docs were previously merged in #535, then reverted in #547.

DO NOT MERGE until the feature is included in a BAAS release. After release:

- [x] Update release date placeholders to the day of the BAAS release.

### Jira

- https://jira.mongodb.org/browse/DOCSP-31628

### Staged Changes

- [Sync Settings](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31628/sync/configure/sync-settings/#development-mode)

### Review Guidelines


dacharyc (approved):
LGTM, although I'd suggest adding a checkbox to the PR description to help you remember to update the date in ln 1 of the `includes` and ln 78, 80, 84, 86 of sync-settings.txt once the release actually goes out.

source/sync/configure/sync-settings.txt
```diff
@@ -61,15 +61,48 @@ update the backend schema by changing the client model.
...
+   
+   For Apps created before [ADD RELEASE DATE], you need to update your
+   object schema in the App Services UI. For details, see
+   :ref:`<synced-schema-overview>`
+
+To make a breaking change from client code:
+
+#. Change your local object model.
+#. Delete your local realm and data.
+#. Open a new realm with your updated object model.
```
yaseenmongo, line 94: Instead of saying `"new"` should it just be `"Open the realm..."`? It is technically a new realm under the hood, but I think as far as the users are concerned it's the same realm with just a different schema. The "new" terminology could potentially lead to confusion (e.g. "what happens to the old realm" "how do I open a new realm").

I think there is also another step of "Run your client application" to ensure that it connects and applies the changes to the server schema.

yaseenmongo (approved):
A couple small questions, but overall LGTM. Thank you! 🚢 🚢 🚢

source/sync/configure/sync-settings.txt
krollins-mdb: Hmm. I see what you mean. I think just "Open a realm" would work. Though it may be good to make it clear that it is, in fact, not the same realm. This could help reinforce the locally-destructive nature of the process.

And good point about adding another step there. I'll update!

yaseenmongo: Open a new realm also works. I just don't want users to think they have to do something differently. They really just need to do `Realm.delete(config)` before the existing `Realm.open()` they already had.

krollins-mdb:
@yaseenmongo, I added a small prerequisites section. Does it look correct to you? https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31628/sync/configure/sync-settings/#breaking-changes

If so, we can merge this PR and update as the other SDKs release with Core v13.19.0.

source/sync/configure/sync-settings.txt
```diff
@@ -61,15 +61,63 @@ update the backend schema by changing the client model.
...
+Breaking Changes
+````````````````
+
+App Services Apps in Development Mode that were created after September 13, 2022
+can make :ref:`breaking changes <breaking-change-quick-reference>` to synced
+object schemas from client code.
+
+**Prerequisites** 
+
+- App Services App created after September 13, 2022
```
yaseenmongo, line 84:
```suggestion
- App Services App created after September 13, 2023
```

I wish it was 2022 too Kyle. There's a couple other spots also with 2022

yaseenmongo (approved):
LGTM! Thank you for your patience Kyle! Sorry for all the delays 🚢 🚢 🚢

source/sync/configure/sync-settings.txt
krollins-mdb: Oh, dang. Thank you for catching that. You'd think I would've realized it's 2023 by... September. 😆

PR #560, opened 2023-08-18
DOCSP-27752 use_natural_pluralization cannot be changed to false usin…

lindseymoore:
…g realm-cli

### Jira

- https://jira.mongodb.org/browse/DOCSP-27752

### Staged Changes

- [GraphQL Configuration Files](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-27752/reference/config/graphql/#service-configuration)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/reference/config/graphql.txt
+++ b/source/reference/config/graphql.txt
@@ -42,12 +42,14 @@ Service Configuration
 
    * - | ``use_natural_pluralization``
        | Boolean
-     - If ``true``, Atlas App Services uses the common English pluralization of a
-       type name whenever possible.
+     - ``true`` by default on new apps. If ``true``, Atlas App Services uses the common 
+       English pluralization of a type name whenever possible.
        
-       If ``false``, or if App Services cannot determine a natural
-       pluralization, the default plural type is the singular type with an
-       ``"s"`` appended to the end.
+       Can set to ``false`` only when you :admin-api-endpoint:`create a new app with the Admin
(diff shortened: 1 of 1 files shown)


source/reference/config/graphql.txt
```diff
@@ -42,12 +42,13 @@ Service Configuration
 
    * - | ``use_natural_pluralization``
        | Boolean
-     - If ``true``, Atlas App Services uses the common English pluralization of a
-       type name whenever possible.
+     - ``true`` by default on new apps. If ``true``, Atlas App Services uses the common
```
MongoCaleb, line 45: How about this? I think we need to explicitly say here it can't be changed. I'm not sure if I'm changing the meaning:

Defaults to``true`` when creating new apps. You can only set the value to ``false`` when you :admin-api-endpoint:`create a new app with the Admin API endpoint <operation/adminCreateApplication>` and pass the property into the body of the request.  The value cannot be changed to ``false`` once the app is created. 

If ``true``, Atlas App Services uses the common English pluralization of a type name whenever possible. If ``false``, or if App Services cannot determine a natural pluralization, the default plural type is the singular type with an ``"s"`` appended to the end.

MongoCaleb (commented):
Suggested change to make the settings a bit more explicit and to join the "if true" and "if false" statements.

MongoCaleb (approved):
LGTM

PR #561, opened 2023-08-18
(DOCSP-29163) App Services Admin API Auth improvements

nlarew:
### Jira

- (DOCSP-29163) App Services Admin API Auth improvements

### Staged Changes

- [Admin API](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/admin-api-refresh-token/admin/api/v3/)

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -3,30 +3,48 @@ openapi: 3.1.0
 info:
   description: |-
     App Services provides an API for programmatically performing
-    administrative tasks outside of the App Services UI. This includes tasks such
-    as modifying authentication providers, creating rules, and defining
-    functions. Client applications should not use the admin API. Instead,
-    use one of the Client SDKs for tasks like authenticating users, calling
-    functions, and interacting with services.
+    administrative tasks outside of the App Services UI. This includes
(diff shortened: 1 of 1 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -3,40 +3,59 @@ openapi: 3.1.0
...
+    - Functions
+    - Logs
+    - Rules
+    - Schemas
+    - Static Hosting
+    - Triggers
+    - Usage & Billing Metrics
+    - Values & Secrets
+
+    Client applications should not use the admin API. Instead, use one
```
MongoCaleb, line 24: The first time I read this, it felt a little weird (like, here's what you can do, but don't). I know that's not what its saying, but that's how it comes across at first. Perhaps something like a note saying  "The admin API is for performing administrative tasks only. Client apps should always use one of the client SDKs for tasks like authenticating users and calling functions."

MongoCaleb (approved):
LGTM with one suggestion.

source/openapi-admin-v3.yaml
nlarew: Gotcha - took what you wrote with a couple tweaks

PR #565, opened 2023-08-28
(DOCSP-31839): Update queryable fields for upcoming release

dacharyc:
## Pull Request Info

This functionality is pending release, and should not be merged until the App Services release currently targeted for Sept 4.

Before merging:

- [x] Confirm the new Sync configuration object structure from the CLI/Admin API and update the [Enable Sync procedure](https://www.mongodb.com/docs/atlas/app-services/sync/configure/enable-sync/#procedure)
- [x] Update the Sync configuration file reference and table from the [Sync Settings page](https://www.mongodb.com/docs/atlas/app-services/sync/configure/sync-settings/#std-label-appconfig-sync)

I split out a separate ticket to update the Admin API Sync endpoint as it is currently missing FS info entirely.

The corresponding [SDK docs PR is 2977](https://github.com/mongodb/docs-realm/pull/2977).

### Jira

- https://jira.mongodb.org/browse/DOCSP-31839

### Staged Changes

- [Sync Settings/Queryable Fields](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31839/sync/configure/sync-settings/#queryable-fields)
- [Sync Errors](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31839/sync/error-handling/errors/#flexible-sync-errors)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sync/configure/enable-sync.txt
+++ b/source/sync/configure/enable-sync.txt
@@ -124,7 +124,9 @@ Procedure
                  "state": <"enabled" | "disabled">,
                  "client_max_offline_days": <Number>,
                  "is_recovery_mode_disabled": <Boolean>,
-                 "queryable_fields_names": <Array of String Field Names>
+                 "indexed_queryable_fields_names": <Array of String Field Names>,
+                 "queryable_fields_names": <Array of String Field Names>,
+                 "collection_queryable_fields_names": <Map[String][]String>
                }
                
             For details, refer to the :ref:`sync-configuration-reference`.
@@ -190,9 +192,11 @@ Procedure
                  "flexible_sync": {
                    "state": "enabled",
                    "database_name": "<Name of Database>",
-                   "queryable_fields_names": <Array of String Field Names>,
                    "client_max_offline_days": <Number>,
-                   "is_recovery_mode_disabled": <Boolean>
+                   "is_recovery_mode_disabled": <Boolean>,
+                   "indexed_queryable_fields_names": <Array of String Field Names>,
+                   "queryable_fields_names": <Array of String Field Names>,
+                   "collection_queryable_fields_names": <Map[String][]String>
                  }
                  ...
                }
--- a/source/sync/configure/sync-settings.txt
+++ b/source/sync/configure/sync-settings.txt
@@ -405,6 +405,10 @@ Sync (recommended). If you are using the older Partition-Based Sync, refer to
      "queryable_fields_names": [
        <Array of String Field Names>
      ],
+     "indexed_queryable_fields_names": [
+      <Array of String Field Names>
+     ],
+     "collection_queryable_fields_names": <Map[String][]String>
      "permissions": "<Deprecated, Do Not Use>"
    }
 
@@ -487,8 +491,28 @@ Sync Config Object
    * - | ``queryable_fields_names``
        | Array<String>
      
-     - The :ref:`names of the fields <queryable-fields>` that your client 
-       application can query to determine which data to synchronize.
+     - The :ref:`names of the global queryable fields <queryable-fields>` 
+       that your client application can query to determine which data to 
+       synchronize. These fields are queryable across all synced collections.
+
+   * - | ``indexed_queryable_fields_names``
+       | Array<String>
+     
+     - The :ref:`names of the indexed queryable fields <fs-indexed-queryable-fields>` 
+       that your client application can query to determine which data to 
+       synchronize. Your app can have one indexed queryable field, and it
+       must be present and be the same 
+       :ref:`eligible field type <flexible-sync-eligible-field-types>`
+       in every collection.
+
+   * - | ``collection_queryable_fields_names``
+       | <Map[String][]String>
+     
+     - The :ref:`names of the collection queryable fields <queryable-field-scopes>` 
+       that your client application can query to determine which data to 
+       synchronize. Collection queryable fields are scoped to the specified
+       collection. Scoping a queryable field to a specific collection reduces
+       the amount of backing Atlas storage required to store Sync metadata.
 
    * - | ``last_disabled``
        | Number
--- a/source/sync/configure/sync-settings.txt
+++ b/source/sync/configure/sync-settings.txt
@@ -276,14 +276,17 @@ example, ``user_id == 641374b03725038381d2e1fb`` or ``store_id IN {1,2,3}``.
 You can optionally include an ``AND`` comparison as long as the indexed
 queryable field is directly compared against a constant using ``==`` or ``IN``
 at least once. For example, ``store_id IN {1,2,3} AND region=="Northeast"``
-or ``store_id == 1 AND active_promotions < 5``.
+or ``store_id == 1 AND (active_promotions < 5 OR num_employees < 10)``.
 
 *Invalid* Flexible Sync queries on an indexed queryable field include queries 
 where:
 
 - The indexed queryable field does not use ``AND`` with the rest of the query.
   For example ``store_id IN {1,2,3} OR region=="Northeast"`` is invalid
-  because it uses ``OR`` instead of ``AND``.
+  because it uses ``OR`` instead of ``AND``.  Similarly, 
+  ``store_id == 1 AND active_promotions < 5 OR num_employees < 10`` is invalid
+  because the ``AND`` only applies to the term next to it, not the entire
+  query.
 - The indexed queryable field is not used in an equality operator. For example
   ``store_id > 2 AND region=="Northeast"`` is invalid because it uses only 
   the ``>`` operator with the indexed queryable field and does not have an 
--- a/source/sync/configure/sync-settings.txt
+++ b/source/sync/configure/sync-settings.txt
@@ -261,9 +261,8 @@ You can still change this value directly in the Atlas database.
 
 .. warning::
 
-   Avoid changing an object's indexed queryable field value when 
-   possible. It may result in dropping other updates on the same 
-   object.
+   Changing an object's indexed queryable field value through Atlas may 
+   overwrite concurrent client updates to the object.
 
 .. _fs-indexed-queryable-fields-valid-client-side-queries:
 
@@ -298,8 +297,9 @@ where:
 Consequences of Adding or Removing Queryable Fields
 ```````````````````````````````````````````````````
 
-You can :ref:`update your Sync configuration <alter-your-sync-configuration>` to
-add or remove queryable field names, but be aware of the following:
+You can :ref:`update your Sync configuration <alter-your-sync-configuration>` 
+to add or remove queryable field names while Sync is enabled, but be aware 
+of the following:
 
 When you add a queryable field, devices can only sync on that field once the
 device has caught up to the point in time in :ref:`Device Sync History
@@ -310,6 +310,11 @@ their Device Sync session dropped and must perform a :ref:`client reset
 <client-resets>`. Clients not using the removed field won't receive any errors.
 To avoid triggering a client reset when you remove the queryable field, you
 should first remove usage of that field on the client-side.
+
+If you :ref:`terminate Sync <terminating-realm-sync>` before adding or 
+removing queryable fields, these considerations do not apply. However, 
+terminating Sync does trigger a :ref:`client reset <client-resets>` for
+any client that has Synced with your App.
 
 Permissions
 ~~~~~~~~~~~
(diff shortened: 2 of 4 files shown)


source/sync/configure/sync-settings.txt
```diff
@@ -173,11 +201,64 @@ Performance and Storage
 Each queryable field adds additional metadata storage to your Atlas cluster and
 may lead to degraded write performance. You should have as few queryable fields
 as needed by your application. A good rule of thumb is to have at most 10
-queryable fields.
+queryable fields. If you need to reduce storage usage or improve performance,
```
sean-brandenburg, line 204: I think this rule of thumb can be updated to something like "10 queryable fields applying to any single collection" to indicate that between an apps global+collection queryable fields they should aim for less than 10 to apply to any collection

source/sync/configure/sync-settings.txt
```diff
@@ -173,11 +201,64 @@ Performance and Storage
...
+storage when using Flexible Sync <optimizing-performance-and-storage-flexible-sync>`.
+
+.. _fs-indexed-queryable-fields:
+
+Indexed Queryable Fields
+++++++++++++++++++++++++
+
+You can improve performance for certain types of workloads by adding an
+indexed queryable field. An **indexed queryable field** is a 
+**global queryable field** that is indexed in the history store, providing
```
sean-brandenburg, line 218: I think `history store` is maybe leaking a bit implementation detail than is needed here. I think something like this is probably enough information

`An **indexed queryable field** is a special **global queryable field** that can be queried on more efficiently, providing improved Sync performance.`

source/sync/configure/sync-settings.txt
```diff
@@ -173,11 +201,64 @@ Performance and Storage
 Each queryable field adds additional metadata storage to your Atlas cluster and
 may lead to degraded write performance. You should have as few queryable fields
 as needed by your application. A good rule of thumb is to have at most 10
-queryable fields.
+queryable fields. If you need to reduce storage usage or improve performance, 
+you can remove unneeded queryable fields from your App. However, be aware
+of the consequences of adding or removing queryable fields detailed below.
```
lindseymoore, line 206:
```suggestion
of the consequences of adding or removing queryable fields detailed below in the section, "Consequences of Adding or Removing Queryable Fields".
```
So reader knows where to expect the information. Think we mostly said no to linking to sections within pages, so including the title name will at least allow reader to use TOC.

source/sync/configure/sync-settings.txt
```diff
@@ -124,26 +124,51 @@ Queryable Fields
...
+.. example::
+
+   In a to-do list app, you might set ``assignee`` or ``owner`` 
+   as queryable fields. On the client side, you can then query for tasks 
+   whose ``assignee`` or ``owner`` matches the logged-in user. 
+
+Queryable Field Scopes
+``````````````````````
+
+Queryable fields apply to a scope you designate when you configure them. The
```
lindseymoore, line 139:
```suggestion
Queryable fields apply to a scope of collections you designate when you configure the fields. The
```
Suggestion: Just for clarity up front as to what the scope applies to.

lindseymoore (approved):
Small comments, LGTM!

source/sync/configure/sync-settings.txt
```diff
@@ -173,11 +201,64 @@ Performance and Storage
 Each queryable field adds additional metadata storage to your Atlas cluster and
 may lead to degraded write performance. You should have as few queryable fields
 as needed by your application. A good rule of thumb is to have at most 10
-queryable fields.
+queryable fields. If you need to reduce storage usage or improve performance, 
+you can remove unneeded queryable fields from your App. However, be aware
+of the consequences of adding or removing queryable fields detailed below.
+
+For additional considerations, refer to :ref:`optimizing performance and
```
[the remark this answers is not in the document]
dacharyc: Good note! I've added more details and an example here, and also built out an example on the "Optimize Sync Storage in Atlas" page.

source/sync/configure/sync-settings.txt
```diff
@@ -124,26 +124,51 @@ Queryable Fields
...
 
 You can automatically specify queryable fields by enabling :ref:`Development
 Mode <development-mode>`. Development Mode allows App Services to automatically
-mark fields as queryable as they are used.
-
-Queryable fields apply across all collections in an App's Schema. You
-can use :ref:`rules and permissions <flexible-sync-rules-and-permissions>` to
-configure more granular access control on individual collections.
+mark fields as queryable as they are used. Queryable fields that you add
+through Development Mode become **collection queryable fields**.
```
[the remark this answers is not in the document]
dacharyc: Updated

source/sync/configure/enable-sync.txt
```diff
@@ -124,7 +124,9 @@ Procedure
                  "state": <"enabled" | "disabled">,
                  "client_max_offline_days": <Number>,
                  "is_recovery_mode_disabled": <Boolean>,
-                 "queryable_fields_names": <Array of String Field Names>
+                 "indexed_queryable_fields_names": <Array of String Field Names>,
```
sean-brandenburg, line 127: Just as a note, it might be a bit confusing that we only allow one indexed_queryable_field but that this is an array (This is to allow us to potentially support more than 1 IQF in the future). 

Would it be possible to add a note here that this array should contain maximum 1 element (here or in the other section describing the config fields)?

source/sync/configure/sync-settings.txt
```diff
@@ -172,12 +202,98 @@ Performance and Storage
...
+If you have a field that you only want to query in one collection, but it 
+is configured as a *global queryable field*, this unnecessarily consumes 
+Atlas storage space. For example, if you have a ``user`` field in every 
+collection, but you only use it for Sync queries in one collection, scoping 
+that as a *collection queryable field* reduces storage requirements. Reducing
+the scope means that Sync does not have to maintain metadata for that field
+for the other collections where you are not querying on the ``user`` field.
+
+If you need to reduce storage usage or improve performance, you can remove 
+unneeded queryable fields from your App. However, be aware of the consequences
```
sean-brandenburg, line 223: The 'consequences of adding/removing a queryable field here' only apply while sync is running. Would you mind making a note about this here?

source/sync/configure/sync-settings.txt
```diff
@@ -172,12 +202,98 @@ Performance and Storage
...
+if your indexed queryable field is ``store_id``, the client cannot change 
+this value directly. Changing it from the client is not supported because 
+it may conflict with other updates made to the object in the same timeframe. 
+
+You can still change this value directly in the Atlas database. 
+
+.. warning::
+
+   Avoid changing an object's indexed queryable field value when 
+   possible. It may result in dropping other updates on the same
```
sean-brandenburg, line 265: This warning might be a bit scarier than it needs to be. This is not something that we explicitly want user's to avoid doing if it's necessary for their use case, but they should be aware that their write may clobber other updates.

What do you think about a warning like this?

`
Changing an object's indexed queryable field value through Atlas may overwrite concurrent client updates to the object.
`

source/sync/configure/sync-settings.txt
```diff
@@ -372,8 +491,28 @@ Sync Config Object
...
-     - The :ref:`names of the fields <queryable-fields>` that your client 
-       application can query to determine which data to synchronize.
+     - The :ref:`names of the global queryable fields <queryable-fields>` 
+       that your client application can query to determine which data to 
+       synchronize. These fields are queryable across all synced collections.
+
+   * - | ``indexed_queryable_fields_names``
+       | Array<String>
+     
+     - The :ref:`names of the indexed queryable fields <fs-indexed-queryable-fields>`
```
sean-brandenburg, line 501: [nit] names -> name

We only allow one IQF at the moment even though this is an array. This is to allow us flexibility in the future for potentially supporting more than 1  IQF

source/sync/go-to-production/optimize-sync-atlas-usage.txt
```diff
@@ -204,11 +204,26 @@ which collections.
...
+   collections, only one queryable field is used.
+
+   Alternately, you may have ``owner_id`` in multiple collections, but only
+   need to query on it in one collection. In this case, you might make 
+   ``owner_id`` a **collection queryable field**. This means Sync only has to
+   maintain metadata about this field for one collection, instead of storing
+   metadata for all of the collections where you're not querying on this 
+   field.
+
+   Finally, for simple equality queries, such as ``owner_id == user.id``,
```
sean-brandenburg, line 220: I think the simple equality note here is not entirely accurate. 

The use case that IQFs make sense in is the same use case as partition based sync where devices want to query on one partition of their data 

Ex. they only care about data for a small subset of (or just one) store/user/etc.

sean-brandenburg (commented):
Thanks for the quick turn around on this! This is looking great, just a few more minor comments

source/sync/configure/enable-sync.txt
dacharyc: It's our convention to list only the type in the config object example, but I've got more details in the table that describes each of the fields: https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-31839/sync/configure/sync-settings/#sync-config-object I'll rework this a bit to be even more explicit.

source/sync/configure/sync-settings.txt
dacharyc: This is a good improvement - thanks! Updating to use this wording.

source/sync/configure/sync-settings.txt
dacharyc: Updated to `name` and added more explicit info that this array may contain one element.

source/sync/go-to-production/optimize-sync-atlas-usage.txt
dacharyc: Gotcha. I've rewritten this section a bit to be more explicit about the use case. Let me know if this needs further refinement!

source/sync/go-to-production/optimize-sync-atlas-usage.txt
```diff
@@ -204,11 +204,30 @@ which collections.
...
+   need to query on it in one collection. In this case, you might make 
+   ``owner_id`` a **collection queryable field**. This means Sync only has to
+   maintain metadata about this field for one collection, instead of storing
+   metadata for all of the collections where you're not querying on this 
+   field.
+
+   Finally, for Apps where devices want to query one specific facet of the 
+   data, such as ``owner_id == user.id``, you may want to designate the 
+   field an **indexed queryable field**. Indexed queryable fields provide 
+   more efficient storage and performance for Apps where the client only
```
sean-brandenburg, line 223: 2 small notes on this sentence
- IQFs increase performance but don't  result in more efficient storage (in fact it will increase the amount of storage used since we're creating an index)
- The last bit of this sentence reads a bit funny to me. Thoughts on something like this?

`Indexed queryable fields provide more efficient performance for Apps where the client only needs to sync on a small subset of their data - a group of stores or a single user, for example.`

source/sync/configure/enable-sync.txt
sean-brandenburg: Cool, this copy in the table seems explicit enough to me. Thanks!

source/sync/configure/sync-settings.txt
```diff
@@ -372,8 +499,29 @@ Sync Config Object
...
-     - The :ref:`names of the fields <queryable-fields>` that your client 
-       application can query to determine which data to synchronize.
+     - The :ref:`names of the global queryable fields <queryable-fields>` 
+       that your client application can query to determine which data to 
+       synchronize. These fields are queryable across all synced collections.
+
+   * - | ``indexed_queryable_fields_names``
+       | Array<String>
+     
+     - The :ref:`name of the indexed queryable field <fs-indexed-queryable-fields>`
```
sean-brandenburg, line 509: Also probably worth noting here that the entry must also occur in `queryable_fields_names` since it needs to be a global queryable field

sean-brandenburg (approved):
Just one comment but otherwise looks good to me! Thanks for the quick iteration on this

dacharyc:
> Is there a release cycle for docs or can I just ping you when these features are released to release these docs at the same time?

@sean-brandenburg This varies depending on the docs set and whether or not we're approaching a big event and implement a code freeze. Generally, though, we publish ad-hoc and I'm happy to release these docs whenever this release lands. Just let me know!

PR #574, opened 2023-09-06
DOCSP-32483 programmatic tagging script for app services repo

sarahemlin:
### Jira

- https://jira.mongodb.org/browse/DOCSP-32483

Hey @dacharyc!

Here's one more set of facet.toml files for your review.  Please merge if they look good to you or let me know if there's some wonkiness or you have questions.

Thank you!
Sarah

### DIFF
--- a/source/examples/generated/flutter/tutorial/facets.toml
+++ b/source/examples/generated/flutter/tutorial/facets.toml
@@ -1,0 +1,7 @@
+[[facets]]
+category = "genre"
+value = "tutorial"
+
+[[facets]]
+category = "programming_language"
+value = "dart"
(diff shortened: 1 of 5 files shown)


source/examples/generated/flutter/tutorial/facets.toml
dacharyc, line 1: I think when we spoke on the call the other day, we decided `generated` files should not have facets. The same thing applies here as in docs-realm - these files are code examples that are used as `includes` on the docs pages so I think we agreed these should be ignored.

dacharyc (commented):
@sarahemlin - back to you with a Q about whether we should include one of these files. The other values seem very reasonable and I'm happy to merge after we remove the facet on the `generated` directory - if I'm remembering the results of our convo correctly.

source/examples/generated/flutter/tutorial/facets.toml
sarahemlin: Yep, you're right!  Thank you!

sarahemlin:
@dacharyc - generated file deleted; tossing it back to you! :)

PR #582, opened 2023-09-13
Update Sync endpoint to Flexible Sync info

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-32692

### Staged Changes
https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-32692/admin/api/v3/#tag/sync

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -6625,7 +6625,7 @@ components:
               description: The version of the queryable field.
             permissions:
               type: object
-              description: **Deprecated**
+              description: *Deprecated*
             last_disabled:
               type: string
               description: The last time Sync was disabled, represented by the number of seconds since the UNIX epoch.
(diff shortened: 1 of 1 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -1397,30 +1397,24 @@ paths:
       - "$ref": "#/components/parameters/AppId"
       - "$ref": "#/components/parameters/ServiceId"
       - "$ref": "#/components/parameters/IncomingWebhookId"
-  "/groups/{groupId}/apps/{appId}/sync/data":
```
nlarew, line 1400: Is this endpoint still relevant for PBS? If so, should we keep it around and make it clear that FS uses the service config?

nlarew (approved):
Maybe we should still include the PBS info for backwards compatibility, at least for a while?

A couple of small suggestions but overall the new stuff LGTM!

PR #597, opened 2023-10-03
add geopoint

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-33226

### Staged Changes

https://preview-mongodbmongocaleb.gatsbyjs.io/atlas-app-services/DOCSP-33226/schemas/types/#geospatial-data

### DIFF
--- a/source/schemas/types.txt
+++ b/source/schemas/types.txt
@@ -525,14 +525,14 @@ GeoJSON Point
 A GeoJSON Point (GeoPoint) is a single location on the Earth's surface. The GeoPoint 
 schema must contain the following fields:
 
-  - ``type`` (string), which is always set to "Point"
-  - ``coordinates`` (array of doubles). The ``coordinates`` array must contain  
-     at least 2 double values, and may contain a third:
-     
-     - The first double is the longitude of the point, between -180 and 180.
-     - The second is the latitude, between -90 and 90.
-     - The optional third value represents the elevation/altitude of the point, 
-       in meters.
+- ``type`` (string), which is always set to "Point"
+- ``coordinates`` (array of doubles). The ``coordinates`` array must contain  
+    at least 2 double values, and may contain a third:
+    
+    - The first double is the longitude of the point, between -180 and 180.
+    - The second is the latitude, between -90 and 90.
+    - The optional third value represents the elevation/altitude of the point, 
+      in meters.
 
 To store a GeoPoint in Atlas for later querying with :ref:`Atlas Device Sync <sync>`, 
 the following rules also apply:
@@ -565,13 +565,11 @@ property named "location". The highlighted rows show an optional property:
              "items": {
                "bsonType": "double"
              }
-           },
-           "additional_location_property": {
-             "bsonType": "string"
-           },
+           }
+          }
          }
-       }
-     }
+       },
+    "additionalProperties": true
    }
 
 The following code block shows a document that follows the schema. It has a 
@@ -607,7 +605,9 @@ query. The Device Sync server validates every embedded GeoPoint property with
   and the second element (latitude) must fall in the range of [-90, 90].
 
 If a device writes an object that fails validation, it will be reverted with a 
-:ref:`compensating write error <flexible-sync-errors>`. 
+:ref:`compensating write error <flexible-sync-errors>`. If a document that doesn't 
+meet these requirements is written directly to the MongoDB collection, the server 
+marks that document as unsyncable.
 
 Querying
 ````````
--- a/source/schemas/types.txt
+++ b/source/schemas/types.txt
@@ -526,8 +526,9 @@ A GeoJSON Point (GeoPoint) is a single location on the Earth's surface. The GeoP
 schema must contain the following fields:
 
 - ``type`` (string), which is always set to "Point"
+
 - ``coordinates`` (array of doubles). The ``coordinates`` array must contain  
-    at least 2 double values, and may contain a third:
+  at least 2 double values, and may contain a third:
     
     - The first double is the longitude of the point, between -180 and 180.
     - The second is the latitude, between -90 and 90.
@@ -544,11 +545,10 @@ the following rules also apply:
 - Additional properties in the GeoPoint object are also allowed.
  
 The following is the Device Sync schema of an object with an embedded GeoPoint 
-property named "location". The highlighted rows show an optional property:
+property named "location":
 
 .. code-block:: json
    :copyable: false
-   :emphasize-lines: 19-21
 
    {
      "title": "MyObject",
@@ -588,7 +588,7 @@ and the GeoPoint object contain an extra property.
        "coordinates":[-122.4, 48.12, 23.0],
        "signal_strength":-118",
        "is_public": true
-       },
+      },
      "is_verified": true
    }
 
@@ -630,7 +630,3 @@ For SDK-specific details about querying against geospatial data, see:
 - :ref:`Geospatial - Node.js SDK <node-data-types-geospatial>`
 - :ref:`Geospatial - React Native SDK <react-native-data-types-geospatial>`
 
-.. note::
-
-   Atlas does not support creating an index on Geospatial fields, so querying 
-   large collections may reduce performance.
--- a/source/schemas/types.txt
+++ b/source/schemas/types.txt
@@ -515,9 +515,11 @@ with the ``title`` field set to the embedded object's type name:
 Geospatial Data
 ---------------
 
-Geospatial data describes points and other data on the earth's surface. Realm 
-does not have built-in geospatial types. If you are not using Device Sync, you 
-can query your geospatial data with regular 
+Geospatial data describes points and other data on the earth's surface. App 
+Services does not have built-in geospatial types. Instead, you model geographic
+data using standard GeoJSON objects.
+
+If you are not using Device Sync, you can query your geospatial data with regular 
 :manual:`geospatial operators </geospatial-queries>`. However, if you are 
 using Device Sync, there are additional requirements in how you store and 
 query the data.
@@ -526,20 +528,26 @@ GeoJSON Point
 ~~~~~~~~~~~~~
 
 A GeoJSON Point (GeoPoint) is a single location on the Earth's surface. The GeoPoint 
-schema must contain the following **required** fields:
+schema must a **required** ``type`` field, of type string, which is always set 
+to "Point".
 
-- ``type`` (string), which is always set to "Point"
-
-- ``coordinates`` (array of doubles). The ``coordinates`` array must contain  
-  at least 2 double values, and may contain a third:
+You must also provide a ``coordinates`` field, which is an array of doubles. 
+The ``coordinates`` array must contain  at least 2 double values, and may 
+contain a third:
     
-  - The first double is the longitude of the point, between -180 and 180.
-  - The second is the latitude, between -90 and 90.
-  - The optional third value represents the elevation/altitude of the point, 
-    in meters. Atlas ignores this value when performing queries. 
+- The first double is the longitude of the point, between -180 and 180.
+- The second is the latitude, between -90 and 90.
+- The optional third value represents the elevation/altitude of the point, 
+  in meters. Atlas ignores this value when performing queries. 
+
+.. note::
+
+   You cannot set an array as required in a Device Sync schema, so the Sync 
+   server checks that the coordinates field is present and meets the 
+   requirements.
 
 To store a GeoPoint in Atlas for later querying with :ref:`Atlas Device Sync <sync>`, 
-the following rules also apply:
+the following rules also apply to the GeoPoint object:
 
 - It must be of type ``object``
 - It must be embedded within another type
@@ -580,14 +588,13 @@ and the GeoPoint object contain an extra property.
 
 .. code-block:: json
    :copyable: false
-   :emphasize-lines: 4-9
+   :emphasize-lines: 3-6
 
    {
      "_id": { "$oid": "65039d09fe4e46dddee31a3f" },
      "location": {
        "type": "Point",
-       "coordinates": [-122.4, 48.12, 23.0],
-       "optional_property": true
+       "coordinates": [-122.4, 48.12, 23.0]
       },
      "another_optional_property": "geopoint name"
    }
@@ -597,12 +604,7 @@ GeoJSON Points and Device Sync Queries
 
 GeoPoints are the only form of geospatial data that Device Sync can 
 query. The Device Sync server validates every embedded GeoPoint property with 
-2 rules:
-
-- The "type" field can only be set to the string "Point" (case-sensitive), 
-- The "coordinates" field can only be set to an array of doubles. The first element 
-  in this array (longitude) must fall in the range of [-180, 180], 
-  and the second element (latitude) must fall in the range of [-90, 90].
+the rules outlined above.
 
 If a device writes an object that fails validation, it will be reverted with a 
 :ref:`compensating write error <flexible-sync-errors>`. If a document that doesn't 
@@ -612,10 +614,9 @@ marks that document as unsyncable.
 Querying
 ````````
 You can query a GeoPoint using :ref:`Realm Query Language (RQL) <rql>`, 
-using the :manual:`$geoWithin </reference/operator/query/geoWithin>` to do so.
-Although GeoPoints are the only stored type you can query, your query will often 
-contain other geospatial types. For example, this partial query looks for all points 
-within the specified GeoCircle:
+using ``$geoWithin`` to do so. Although GeoPoints are the only stored type you 
+can query, your query may contain other geospatial types. For example, 
+this partial query looks for all points within the specified GeoCircle:
 
 .. code-block:: json
    :copyable: false
--- a/source/schemas/types.txt
+++ b/source/schemas/types.txt
@@ -549,9 +549,9 @@ contain a third:
 To store a GeoPoint in Atlas for later querying with :ref:`Atlas Device Sync <sync>`, 
 the following rules also apply to the GeoPoint object:
 
-- It must be of type ``object``
-- It must be embedded within another type
-- Additional properties are allowed within the ``coordinates`` array, as long as 
+- It must be of type ``object``.
+- It must be embedded within another type.
+- Additional values are allowed within the ``coordinates`` array, as long as 
   they are doubles.
 - Additional properties in the GeoPoint object are also allowed.
  
@@ -564,6 +564,9 @@ property named "location":
    {
      "title": "MyObject",
      "properties": {
+        "_id": {
+          "bsonType": "objectId"
+        },
         "location": {
           "bsonType": "object",
           "required": [ "type" ],
@@ -583,8 +586,7 @@ property named "location":
    }
 
 The following code block shows a document that follows the schema. It has a 
-single embedded GeoPoint object named "location". Both the containing object 
-and the GeoPoint object contain an extra property.
+single embedded GeoPoint object named "location".
 
 .. code-block:: json
    :copyable: false
@@ -595,8 +597,7 @@ and the GeoPoint object contain an extra property.
      "location": {
        "type": "Point",
        "coordinates": [-122.4, 48.12, 23.0]
-      },
-     "another_optional_property": "geopoint name"
+      }
    }
 
 GeoJSON Points and Device Sync Queries
@@ -609,12 +610,12 @@ the rules outlined above.
 If a device writes an object that fails validation, it will be reverted with a 
 :ref:`compensating write error <flexible-sync-errors>`. If a document that doesn't 
 meet these requirements is written directly to the MongoDB collection, the server 
-marks that document as unsyncable.
+marks that document as :ref:`unsyncable <unsynced-documents>`.
 
 Querying
 ````````
 You can query a GeoPoint using :ref:`Realm Query Language (RQL) <rql>`, 
-using ``$geoWithin`` to do so. Although GeoPoints are the only stored type you 
+using ``geoWithin`` to do so. Although GeoPoints are the only stored type you 
 can query, your query may contain other geospatial types. For example, 
 this partial query looks for all points within the specified GeoCircle:
 
@@ -629,4 +630,3 @@ For SDK-specific details about querying against geospatial data, see:
 - :ref:`Geospatial - Kotlin SDK <kotlin-geospatial>`
 - :ref:`Geospatial - Node.js SDK <node-data-types-geospatial>`
 - :ref:`Geospatial - React Native SDK <react-native-data-types-geospatial>`
-
--- a/source/schemas/types.txt
+++ b/source/schemas/types.txt
@@ -509,3 +509,74 @@ with the ``title`` field set to the embedded object's type name:
             }
       }
    }
+
+.. _geospatial
+
+
+Geospatial Data
+~~~~~~~~~~~~~~~
+
+Geospatial data describes points and other data on the earth's surface. App 
+Services supports defining GEOPoint data in a schema. 
+
+GeoJSON Point
+`````````````
+A GeoJSON point is a single location on the Earth's surface, described by its 
+longitude and latitude. To ensure wide compatibility, App Services stores these 
+points in accordance to the `GeoJSON spec <https://datatracker.ietf.org/doc/html/rfc7946>`_, 
+which has additional requirements. Specifically, you must apply the following 
+rules to the GeoPoint:
+
+- It must be of type ``object``
+- It must be embedded within another type
+- It must contain the following fields:
+
+  - ``type``, of type ``string``, which is always set to "Point"
+  - ``coordinates``, of type ``array``. The ``coordinates`` array contains 
+    exactly 2 double values. The first double is the longitude of the point and 
+    the second is the latitude.
+
+- Additional properties are allowed.
+ 
+ The following is the schema of a GeoPoint:
+
+.. code-block:: json
+
+{
+  "title": "MyObject",
+  "properties": {
+    "location": {
+      "bsonType": "object",
+      "required": [ "type" ],
+      "properties": {
+        "type": {
+          "bsonType": "string",
+        },
+        "coordinates": {
+          "bsonType": "array",
+          "items": {
+            "bsonType": "double"
+          }
+        }
+      }
+    },
+    "another_property": {
+          "bsonType": "double"
+    },
+  }
+}
+
+You can query geospatial data when using :ref:`Flexible Sync <flexible-sync>`. 
+No additional configuration is required. 
+
+To learn about querying against geospatial data, see:
+
+- :ref:`Geospatial - .NET SDK <dotnet-geospatials
+- :ref:`Geospatial - Kotlin SDK <kotlin-geospatial
+- :ref:`Geospatial - Node.js SDK <node-data-types-geospatial>
+- :ref:`Geospatial - React Native SDK <react-native-data-types-geospatial
+
+.. note::
+
+   Atlas does not support creating an index on Geospatial fields, so querying 
+   large collections may reduce performance.
--- a/source/schemas/types.txt
+++ b/source/schemas/types.txt
@@ -510,8 +510,7 @@ with the ``title`` field set to the embedded object's type name:
       }
    }
 
-.. _geospatial
-
+.. _geospatial:
 
 Geospatial Data
 ---------------
@@ -524,22 +523,22 @@ GeoJSON Point
 
 A GeoJSON Point is a single location on the Earth's surface, described by its 
 longitude and latitude. To ensure wide compatibility, App Services stores these 
-points in accordance to the `GeoJSON spec <https://datatracker.ietf.org/doc/html/rfc7946>`_, 
-which has additional requirements. Specifically, you must apply the following 
-rules to the GeoPoint:
+points in accordance to the `GeoJSON spec <https://datatracker.ietf.org/doc/html/rfc7946>`_.
+To store a GeoPoint in Atlas, you must follow these rules:
 
 - It must be of type ``object``
 - It must be embedded within another type
 - It must contain the following fields:
 
-  - ``type``, of type ``string``, which is always set to "Point"
+  - ``type``, which is always set to "Point"
   - ``coordinates``, of type ``array``. The ``coordinates`` array contains 
     exactly 2 double values. The first double is the longitude of the point and 
     the second is the latitude.
 
 - Additional properties are allowed.
  
- The following is the schema of a GeoPoint:
+The following is the schema of an object where the "location" property is 
+a GeoPoint:
 
 .. code-block:: json
 
@@ -552,17 +551,20 @@ rules to the GeoPoint:
          "properties": {
            "type": {
              "bsonType": "string",
+             "const": "Point"
            },
            "coordinates": {
              "bsonType": "array",
              "items": {
-               "bsonType": "double"
+               "bsonType": "double",
+                "minItems": 2,
+                "maxItems": 2
              }
            }
          }
        },
-       "another_property": {
-             "bsonType": "double"
+       "additional_location_property": {
+             "bsonType": "string"
        },
      }
    }
--- a/source/schemas/types.txt
+++ b/source/schemas/types.txt
@@ -515,32 +515,40 @@ with the ``title`` field set to the embedded object's type name:
 Geospatial Data
 ---------------
 
-Geospatial data describes points and other data on the earth's surface. App 
-Services supports defining GeoJSON Point data in a schema. 
+Geospatial data describes points and other data on the earth's surface. The 
+Realm SDKs support the creation of numerous types of geospatial data. However, 
+only persisted GeoJSON Point data can be queried. 
 
 GeoJSON Point
 ~~~~~~~~~~~~~
 
-A GeoJSON Point is a single location on the Earth's surface, described by its 
-longitude and latitude. To ensure wide compatibility, App Services stores these 
-points in accordance to the `GeoJSON spec <https://datatracker.ietf.org/doc/html/rfc7946>`_.
-To store a GeoPoint in Atlas, you must follow these rules:
+A GeoJSON Point (GeoPoint) is a single location on the Earth's surface. The GeoPoint 
+schema must contain the following fields:
+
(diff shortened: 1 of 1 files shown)


source/schemas/types.txt
```diff
@@ -509,3 +509,78 @@ with the ``title`` field set to the embedded object's type name:
...
+       "another_property": {
+             "bsonType": "double"
+       },
+     }
+   }
+
+Querying GeoJSON Points
+~~~~~~~~~~~~~~~~~~~~~~~
+
+You can query geospatial points when using :ref:`Flexible Sync <flexible-sync>`.
```
nlarew, line 573: Leads me to wonder how I do geo queries with GraphQL & Data API (and maybe even MQL queries in Functions). I assume the current GeoSpatial project is only scoped for Sync and not the other services?

source/schemas/types.txt
```diff
@@ -509,3 +509,78 @@ with the ``title`` field set to the embedded object's type name:
...
+
+- :ref:`Geospatial - .NET SDK <dotnet-geospatial>`
+- :ref:`Geospatial - Kotlin SDK <kotlin-geospatial>`
+- :ref:`Geospatial - Node.js SDK <node-data-types-geospatial>`
+- :ref:`Geospatial - React Native SDK <react-native-data-types-geospatial>`
+
+.. note::
+
+   Atlas does not support creating an index on Geospatial fields, so querying 
+   large collections may reduce performance.
```
nlarew, line 586: This just sounds... wrong? Can I not create a 2dsphere index in Atlas?

source/schemas/types.txt
```diff
@@ -509,3 +509,78 @@ with the ``title`` field set to the embedded object's type name:
...
+         "required": [ "type" ],
+         "properties": {
+           "type": {
+             "bsonType": "string",
+           },
+           "coordinates": {
+             "bsonType": "array",
+             "items": {
+               "bsonType": "double"
+             }
```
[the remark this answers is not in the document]
MongoCaleb: clever!

source/schemas/types.txt
MongoCaleb: Correct - FlexSync only

source/schemas/types.txt
MongoCaleb: https://jira.mongodb.org/browse/SERVER-80402

source/schemas/types.txt
```diff
@@ -509,3 +509,80 @@ with the ``title`` field set to the embedded object's type name:
...
+Geospatial data describes points and other data on the earth's surface. App 
+Services supports defining GeoJSON Point data in a schema. 
+
+GeoJSON Point
+~~~~~~~~~~~~~
+
+A GeoJSON Point is a single location on the Earth's surface, described by its 
+longitude and latitude. To ensure wide compatibility, App Services stores these 
+points in accordance to the `GeoJSON spec <https://datatracker.ietf.org/doc/html/rfc7946>`_.
+To store a GeoPoint in Atlas, you must follow these rules:
```
nlarew, line 527: nit - these aren't really requirements for storing in Atlas so much as:

- a schema/definition of a GeoJSON Point in general
- requirements for how you use Point objects in your Sync models (i.e. "It must be embedded within another type")

For the "embedded within another type" requirement, I think we can be clearer that the limitation is specifically if you want to query the `Point` field with geospatial RQL operators. You could totally have a regular object type called `Point` that follows this schema and it would save & sync just fine, right?

source/schemas/types.txt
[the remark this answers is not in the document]
nlarew: One wrinkle here - the spec actually supports an optional third element

> A position is an array of numbers.  There MUST be two or more
> elements.  The first two elements are longitude and latitude, or
> easting and northing, precisely in that order and using decimal
> numbers.  _**Altitude or elevation MAY be included as an optional third
> element.**_

Do you know if we have a hard support/no support stance on the third value?

MongoCaleb:
@jonatlee  - can you help clarify the index issue in the thread above?

source/schemas/types.txt
jonatlee: Yeah, the limitation here is that you cannot create a 2dsphere index on an _array_ of Point objects. Flexible sync will index your query in our internal metadata under the hood and because setting that index is unsupported in MongoDB for an array of Point objects, those will be unindexed queries which would result in degraded performance.

source/schemas/types.txt
```diff
@@ -509,3 +509,80 @@ with the ``title`` field set to the embedded object's type name:
...
+             "bsonType": "array",
+             "items": {
+               "bsonType": "double",
+                "minItems": 2,
+                "maxItems": 2
+             }
+           }
+         }
+       },
+       "additional_location_property": {
```
jonatlee, line 566: [nit] I think to illustrate the point that additional properties can be stored within the "location" embedded object and still be considered a GeoJSON Point by the server, we probably want a structure like this:

```
"location": {
     "bsonType": "object",
     "required": [ "type" ],
     "properties": {
         "type": {
             "bsonType": "string"
         },
         "coordinates": {
             "bsonType": "array"
             "items": {
               "bsonType": "double"
             }
         },
         "additional_location_property": {
             "bsonType": "string"
         }
    }
}
```

source/schemas/types.txt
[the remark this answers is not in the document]
jonatlee: The MongoDB server will tolerate values past the first two elements in the array of doubles. It's just that the query will only actually look at the first two.

source/schemas/types.txt
```diff
@@ -509,3 +509,80 @@ with the ``title`` field set to the embedded object's type name:
...
+         "properties": {
+           "type": {
+             "bsonType": "string",
+             "const": "Point"
+           },
+           "coordinates": {
+             "bsonType": "array",
+             "items": {
+               "bsonType": "double",
+                "minItems": 2,
```
jonatlee, line 560: [nit] We actually decided that "minItems" and "maxItems" weren't worth adding into the JSON schema definition, and the only requirement for "coordinates" is that it must have "bsonType": "array", and items "bsonType": "double".

source/schemas/types.txt
```diff
@@ -509,3 +509,80 @@ with the ``title`` field set to the embedded object's type name:
...
+   {
+     "title": "MyObject",
+     "properties": {
+       "location": {
+         "bsonType": "object",
+         "required": [ "type" ],
+         "properties": {
+           "type": {
+             "bsonType": "string",
+             "const": "Point"
```
jonatlee, line 554: [nit] The "const": "Point" is not necessary in order for the server to interpret an embedded object as a GeoJSON Point. I think we should remove it

source/schemas/types.txt
jonatlee: Perhaps it would also be worth giving an example of the minimum JSON schema requirements for a GeoJSON Point versus it with the additional property. 

So this would be an example of an embedded object satisfying the minimum requirements:
```
"location": {
     "bsonType": "object",
     "required": [ "type" ],
     "properties": {
         "type": {
             "bsonType": "string"
         },
         "coordinates": {
             "bsonType": "array"
             "items": {
               "bsonType": "double"
             }
         }
    }
}
```
And then including the above as reference too

source/schemas/types.txt
```diff
@@ -509,3 +509,80 @@ with the ``title`` field set to the embedded object's type name:
...
+     }
+   }
+
+Querying GeoJSON Points
+~~~~~~~~~~~~~~~~~~~~~~~
+
+You can query geospatial points when using :ref:`Flexible Sync <flexible-sync>`. 
+No additional configuration is required. 
+
+To learn about querying against geospatial data, see:
```
jonatlee, line 578: I think it could be cool to have a demo/walkthrough app where it tells you how to setup a Device Sync app from start to finish that is querying on geospatial data. Perhaps out of the scope of this ticket though

source/schemas/types.txt
```diff
@@ -509,3 +509,80 @@ with the ``title`` field set to the embedded object's type name:
...
+       "additional_location_property": {
+             "bsonType": "string"
+       },
+     }
+   }
+
+Querying GeoJSON Points
+~~~~~~~~~~~~~~~~~~~~~~~
+
+You can query geospatial points when using :ref:`Flexible Sync <flexible-sync>`.
```
jonatlee, line 575: [q] Are we migrating the docs to reference "Atlas Device Sync" instead of "Flexible Sync"? Should we also do that here?

source/schemas/types.txt
```diff
@@ -509,3 +509,80 @@ with the ``title`` field set to the embedded object's type name:
...
+             }
+           }
+         }
+       },
+       "additional_location_property": {
+             "bsonType": "string"
+       },
+     }
+   }
+
```
jonatlee, line 571: I think there are more things that we should mention in the sync-specific docs here; especially because it seems unclear to me at the moment whether future geospatial support will be rolled out in Realm at the same time as Atlas Device Sync.
1. `geoWithin` is the only operator supported in querying in Atlas Device Sync
2. GeoJSON Points are the _only_ form of geospatial data that can be queried in Atlas Device Sync. Note that this does not mean that you can't use the `geoWithin` operator against the supported geospatial types in Realm (eg. `location geoWithin GeoCircle([-73.981209, 40.725055], 0.00050464271`). It's just that the field name being operated on in the geoWithin operator (the LHS of the geoWithin operator) can only correspond to a GeoJSON Point property.
3. The sync server will perform GeoJSON Point validation for changes to **_any_** property that matches the JSON schema embedded object structure noted above (note that this is not just a GeoJSON Point property that is queryable, any property in the JSON schema that matches the embedded object structure above is subject to this!). The checks will specifically ensure that: (1) the "type" field can only be set to the string "Point" (case-sensitive), and (2) the "coordinates" field can only be set to an array of doubles. The first element in the "coordinates" array (longitude) must fall in the range of [-180, 180], and the second element (latitude) must fall in the range of [-90, 90].
  i. If a **device** makes a write that violates the above constraints, then it will be reverted via a [compensating write](https://www.mongodb.com/docs/atlas/app-services/sync/error-handling/errors/#std-label-flexible-sync-errors). 
  ii. If a **write to MongoDB** violates the above constraints, then the offending document will become [unsyncable](https://www.mongodb.com/docs/atlas/app-services/mongodb/internal-database-usage/#unsynced-documents). 
4. There is now a new option in the "BsonType" Select dropdown in the Rich Schema editor in the Schema Page UI that will automatically fill in the necessary requirements for marking a property as a GeoJSON Point. See [here](https://github.com/10gen/baas-ui/pull/3413) for what that looks like.

source/schemas/types.txt
jonatlee: yeah, I guess it may be nice to introduce what the JSON schema for a GeoJSON Point would look like perhaps before diving into how to structure the MongoDB documents themselves (unless I'm misunderstanding your comment). I think the main thing we need to make clear here is that the Atlas Device Sync will only recognize a property as a GeoJSON Point if the JSON schema meets the proper embedded object structure. The structure of the documents themselves should follow this schema, and are subject to additional constraints (noted below in the comment below / in this doc [here](https://docs.google.com/document/d/10aAwBVwWRl2OOiG8RCXfT_-ejIWe6KmYFtrZgGPW15k/edit#heading=h.fgl0tx27vwoj))

jonatlee (commented):
Let me know if you have any questions about my comments!

source/schemas/types.txt
MongoCaleb: @jonatlee I'm confused; isn't the "type" property required _and_ also must be set to "point"? The spec suggests that's the case.

source/schemas/types.txt
MongoCaleb: Yes, cool and way beyond the scope. :) I have built one, so perhaps at some point I can water it down to use as a sample.

source/schemas/types.txt
MongoCaleb: Yes. I let this one slip through. :(

source/schemas/types.txt
jonatlee: [This](https://docs.google.com/document/d/1mOV04pa_Wl4uC5l6wBooQWyv2OVsBll6gtKaWl-ugGE/edit#bookmark=id.y8wrmys9ft74) is the expected minimum JSON schema structure for defining a GeoPoint property. We originally were going to include "enum": ["Point"] similar to the above but we decided that it wasn't necessary. The server performs validation on top of the JSON schema validation that ensures that the "type" field will only ever get set to "Point". That is what 3.) in this comment https://github.com/mongodb/docs-app-services/pull/597#discussion_r1347878102

source/schemas/types.txt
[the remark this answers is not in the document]
MongoCaleb: @nlarew, I changed the `maxItems` to 3, and updated the preceding paragraph to mention the optional elevation double.

source/schemas/types.txt
MongoCaleb: I'm still a bit confused by the intricacies here. You can define a 2dsphere index on documents that contain a single GeoPoint (as per [this doc](https://www.mongodb.com/docs/manual/core/indexes/index-types/geospatial/2dsphere/create/)). But Flexible Sync...does what?...that causes those queries to be un-indexed?

source/schemas/types.txt
MongoCaleb: @jonatlee  So the schema doesn't enforce the presence of the "type" property or it's value as "Point"?

source/schemas/types.txt
MongoCaleb: @jonatlee one last question on the schema: "type" is the only required property? Not "coordinates", too?

source/schemas/types.txt
jonatlee: Yeah, for queries on an array of Point objects, they will be unindexed; however, for queries on a single Point object, they will be indexed with the 2dsphere index.

source/schemas/types.txt
jonatlee: "type" is required -- it needs to be explicitly included in the required fields list in order for the embedded object to be treated as a Point by the server. 

Outside of any schema validation, the value of "type" is checked by the server on properties that have been marked as a GeoPoint as writes are coming in to ensure that it never gets set to anything other than "Point".

If "type" gets set to something other than "Point" by a mobile device, the server will issue a compensating write to the client which effectively reverts the write. 

If "type" gets set to something other than "Point" due to a change directly to MongoDB, the server will mark the written document as unsyncable.

source/schemas/types.txt
MongoCaleb: @jonatlee - confused on point 2. Are you saying any embedded object that "looks like" a GeoPoint is validated? And if so, what does "look like" mean, because if it doesn't meet all the requirements of a GeoPoint, does it get skipped by the validator, or fail validation? How do we draw the line there?

source/schemas/types.txt
jonatlee: Arrays must be optional in order to be sync compatible, so "coordinates" should not be specified as part of the required fields lists. The server will ensure that writes to "coordinates" will always leave it in a valid state though (array of doubles with length >= 2, longitude is in the range [-180, 180] and latitude is in the range [-90, 90])

source/schemas/types.txt
nlarew: We should probably do [`additionalProperties: true`](https://json-schema.org/understanding-json-schema/reference/object#additionalproperties) instead of using a synthetic example field in the schema.

source/schemas/types.txt
jonatlee: Yes, any property that is considered to be a GeoPoint according to the server's understanding of what a GeoPoint is is subject to the validation. If it does not meet all the requirements for specifying a GeoPoint, then no GeoPoint validation will come into play for that property. The line is drawn in that the following minimum requirements must be met in order for it to be considered a GeoPoint:
1. The property is an embedded object
2. The property contains a required field, "type", that is a string
3. The property contains a field, "coordinates", that must be an array of doubles

For example, the "location" field in the example below would fulfill these minimum requirements and be considered a GeoPoint by the server
```
{
    "title": "myTable",
    "required": ["_id"]
    "properties": {
        "_id": { "bsonType": "objectId" },
        "location": {
            "bsonType": "object",
            "required": [ "type" ],
            "properties": {
                "type": {
                    "bsonType": "string"
                },
                "coordinates": {
                    "bsonType": "array",
                    "items": {
                        "bsonType": "double"
                    }
                }
            }
        }
    }
}
```

source/schemas/types.txt
jonatlee: I forget if this is the case, but we might treat the object as a dictionary internally if there is both additionalProperties and properties specified? https://www.mongodb.com/docs/atlas/app-services/schemas/types/#dictionary-of-mixed-bson-types

source/schemas/types.txt
MongoCaleb: @jonatlee My question wasn't clear. I understand what is required to pass validation, but what I don't understand is this:
> any property that is considered to be a GeoPoint according to the server's understanding of what a GeoPoint is is subject to the validation

So the server understands a Geopoint has an embedded object with the above requirements, so in order for it to identify a Geopoint, it has already validated that's what it is? 

Or is it: "I see an embedded object with a 'type' and 'coordinates' properties, but 'type' isn't set to 'Point' (and/or 'coordinates isn't an array of >=2 doubles)"?

jonatlee: The JSON schema requirements for a GeoPoint exists to answer the question of "should we run the GeoPoint validation checks against this property" whenever a write is made by a mobile client or directly to MongoDB. 

On each write, we look at the properties in the updated document. Any property in the updated document is cross-referenced with its JSON schema structure to see if it is a GeoPoint, and if it is one, then we will perform the GeoPoint validation checks on the updated state of the document. This is what allows us to ensure things like "type" only being allowed to be set to "Point" and "coordinates" only containing a list of doubles with valid longitude/latitude values outside of the JSON schema definition. 

Let me know if that makes sense. Also happy to chat over Slack about this if that's easier

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+You can query a GeoPoint using :ref:`Realm Query Language (RQL) <rql>`, 
+using the :manual:`$geoWithin </reference/operator/query/geoWithin>` to do so.
+Although GeoPoints are the only stored type you can query, your query will often 
+contain other geospatial types. For example, this partial query looks for all points 
+within the specified GeoCircle:
+
+.. code-block:: json
+   :copyable: false
+
+   location geoWithin GeoCircle([-73.981209, 40.725055], 0.00050464271
```
jonatlee, line 624:
```suggestion
   location geoWithin GeoCircle([-73.981209, 40.725055], 0.00050464271)
```
[nit] Looks like we are missing a closing parenthesis here

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+           "coordinates": {
+             "bsonType": "array",
+             "items": {
+               "bsonType": "double"
+             }
+           }
+          }
+         }
+       },
+    "additionalProperties": true
```
jonatlee, line 572: Also wanted to ping and check if setting "additionalProperties" will lead to this property being interpreted as a dictionary versus an object. Happy to also follow up with my team if no one knows

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+
+The following code block shows a document that follows the schema. It has a 
+single embedded GeoPoint object named "location". Both the containing object 
+and the GeoPoint object contain an extra property.
+
+.. code-block:: json
+   :copyable: false
+   :emphasize-lines: 4-9
+
+   {
```
jonatlee, line 583: [nit] probably worth adding some whitespace here (similar to the spacing in the above code block) to make the code block a bit more readable

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+     "_id":{"$oid":"65039d09fe4e46dddee31a3f"},
+     "carrier":"SuperCheapMobileProvider",
+     "location":{
+       "type":"point",
```
jonatlee, line 587:
```suggestion
       "type": "Point",
```
"Point" is case-sensitive

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+schema must contain the following fields:
+
+- ``type`` (string), which is always set to "Point"
+
+- ``coordinates`` (array of doubles). The ``coordinates`` array must contain  
+  at least 2 double values, and may contain a third:
+    
+    - The first double is the longitude of the point, between -180 and 180.
+    - The second is the latitude, between -90 and 90.
+    - The optional third value represents the elevation/altitude of the point,
```
jonatlee, line 535: Can we mention here that altitude won't actually be taken into account for queries though?

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+               "bsonType": "double"
+             }
+           }
+          }
+         }
+       },
+    "additionalProperties": true
+   }
+
+The following code block shows a document that follows the schema. It has a
```
jonatlee, line 575: [nit] It seems like the example below is inconsistent with the schema above. Perhaps we could align the two?

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+    - The second is the latitude, between -90 and 90.
+    - The optional third value represents the elevation/altitude of the point, 
+      in meters.
+
+To store a GeoPoint in Atlas for later querying with :ref:`Atlas Device Sync <sync>`, 
+the following rules also apply:
+
+- It must be of type ``object``
+- It must be embedded within another type
+- Additional properties are allowed within the ``coordinates`` array, as long as
```
jonatlee, line 543:
```suggestion
- Additional entries are allowed within the ``coordinates`` array, as long as 
```
[nit] I think "entries" or "items" is probably more accurate here. Properties makes me think of object properties

source/schemas/types.txt
```diff
@@ -509,3 +509,80 @@ with the ``title`` field set to the embedded object's type name:
             }
       }
    }
+
+.. _geospatial:
```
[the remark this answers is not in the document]
jonatlee: Wanted to ping this and see if it still makes sense to keep this here

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+Realm SDKs support the creation of numerous types of geospatial data. However, 
+only persisted GeoJSON Point data can be queried. 
+
+GeoJSON Point
+~~~~~~~~~~~~~
+
+A GeoJSON Point (GeoPoint) is a single location on the Earth's surface. The GeoPoint 
+schema must contain the following fields:
+
+- ``type`` (string), which is always set to "Point"
```
jonatlee, line 528: [nit] I'm predicting the most common thing people will miss when they try to use GeoPoints is that "type" is a required field and must be marked as such in the JSON schema in order for the property to be picked up as a GeoPoint. Could we make that explicit here?

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+and the GeoPoint object contain an extra property.
+
+.. code-block:: json
+   :copyable: false
+   :emphasize-lines: 4-9
+
+   {
+     "_id":{"$oid":"65039d09fe4e46dddee31a3f"},
+     "carrier":"SuperCheapMobileProvider",
+     "location":{
```
jonatlee, line 586: I think between if we want to only show one example of a GeoPoint here, I kind of like having just the minimum requirements for a GeoPoint laid out here (ie without the extra properties -- just "type" and "coordinates").

jonatlee (commented):
Looking pretty good, flushing some more comments

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
    }
+
+.. _geospatial:
+
+Geospatial Data
+---------------
+
+Geospatial data describes points and other data on the earth's surface. The 
+Realm SDKs support the creation of numerous types of geospatial data. However, 
+only persisted GeoJSON Point data can be queried.
```
nlarew, line 520: To @jonatlee's point in another comment, perhaps we could clarify here that there is no built-in geospatial BSON type (which is what the top half of this page talks about).

- If you're not using sync then you can treat geospatial data as a regular object and can query it in MongoDB with regular [geospatial operators](https://www.mongodb.com/docs/manual/geospatial-queries/).
- If you're using sync then there's a smaller set of operators (i.e. only 1 right now) and some additional considerations.

source/schemas/types.txt
nlarew, line 572: We need to move this `additionalProperties` field to describe the `locations` embedded object instead of the root object

MongoCaleb: @jonatlee I don't know. If you don't mind checking, I'd appreciate it. How does it change the schema definition, though?

source/schemas/types.txt
MongoCaleb: They are object properties, though, aren't they? We can have additional properties on the object. ?

source/schemas/types.txt
jonatlee: I can check in with the team, but I guess I also don't think it's totally necessary to include here in the example. I suspect most people will just want to know how to specify a GeoJSON point without any additional properties within the point, so having the minimum definition as an example should suffice imo. Let me know if you feel strongly about this though

source/schemas/types.txt
jonatlee: Gotcha -- it might work for local queries, but for synced queries it needs to be "Point", as MongoDB only will accept "Point" I believe

source/schemas/types.txt
jonatlee: Yeah, so just checked on this -- specifying "additionalProperties" in addition to "properties" for an embedded object will make the entire schema sync-incompatible. So let's omit specifying "additionalProperties" in the example

source/schemas/types.txt
MongoCaleb: @jonatlee to be clear, even if the schema doesn't require the coordinates field, it is also required in order to a) be compliant with the geopoint spec, and b) to do geowithin querying. Yes?

source/schemas/types.txt
MongoCaleb: As per Jonathon's comments, I've removed it entirely.

source/schemas/types.txt
nlarew: I think the confusion is from "IN the coordinates array".

Items in an array (i.e. the lat/lon doubles) are array elements/entries/items.

The sibling fields of `coordinates` are object properties.

source/schemas/types.txt
```diff
@@ -509,3 +509,123 @@ with the ``title`` field set to the embedded object's type name:
             }
       }
    }
+
+.. _geospatial:
+
+Geospatial Data
+---------------
+
+Geospatial data describes points and other data on the earth's surface. Realm
```
nlarew, line 518: Let's replace "Realm" with "App Services"

source/schemas/types.txt
```diff
@@ -509,3 +509,123 @@ with the ``title`` field set to the embedded object's type name:
...
       }
    }
+
+.. _geospatial:
+
+Geospatial Data
+---------------
+
+Geospatial data describes points and other data on the earth's surface. Realm 
+does not have built-in geospatial types. If you are not using Device Sync, you
```
nlarew, line 519:
```suggestion
does not have built-in geospatial types. Instead, you can model geographic
data using standard GeoJSON objects.

If you are not using Device Sync, you 
```

source/schemas/types.txt
jonatlee: yep that is correct. The schema can't have "coordinates" as a required field specifically because marking an array property as required in JSON schema will lead to the schema becoming incompatible with sync

source/schemas/types.txt
```diff
@@ -509,3 +509,123 @@ with the ``title`` field set to the embedded object's type name:
...
+      }
+   }
+
+The following code block shows a document that follows the schema. It has a 
+single embedded GeoPoint object named "location". Both the containing object 
+and the GeoPoint object contain an extra property.
+
+.. code-block:: json
+   :copyable: false
+   :emphasize-lines: 4-9
```
jonatlee, line 583: the code highlighting seems a bit off? unless you were intending on highlighting the geopoint properties + the extra property (but maybe we should do 4-8 instead to not highlight the last curly brace?)

source/schemas/types.txt
jonatlee: that being said^, sorry perhaps we should make it clear that it should only be the "type" field marked as required and "coordinates" not as a required in the JSON schema to prevent customers from accidentally creating sync incompatible schemas. it is true that both must be present however for those documents to actually pass validation though. 

as a side note, i was actually pushing for making "coordinates" a required field in the JSON schema and making required array fields sync compatible, but that was rejected 😅. i think the main reason is that it isn't immediately clear what "required" means for an array (ie, array length > 0, array is non-nil, etc)

source/schemas/types.txt
jonatlee: Got it. I get that we have it here for demonstrative purposes, but i do think for most use cases people will only want to have "type"/"coordinates" specified in the geopoint embedded structure (let me know if you disagree though). That was the main reason why I was proposing keeping it minimal.

I'm on board with having an additional property at the same level as `"location"` in this example (`"another_optional_property"`), but still feel like it's fine to keep the geopoint properties as just "type"/"coordinates"

source/schemas/types.txt
nlarew, line 583:
```suggestion
   :emphasize-lines: 3-8
```
Screenshot 2023-10-16 at 3 34 26 PM

jonatlee (commented):
Looks great! Thanks for the patience/hard work on this, just some last comments before approving (also I don't think I have access to resolve any of the outdated comment threads -- feel free to resolve them)

source/schemas/types.txt
```diff
@@ -509,3 +509,123 @@ with the ``title`` field set to the embedded object's type name:
...
+
+If a device writes an object that fails validation, it will be reverted with a 
+:ref:`compensating write error <flexible-sync-errors>`. If a document that doesn't 
+meet these requirements is written directly to the MongoDB collection, the server 
+marks that document as unsyncable.
+
+Querying
+````````
+You can query a GeoPoint using :ref:`Realm Query Language (RQL) <rql>`, 
+using the :manual:`$geoWithin </reference/operator/query/geoWithin>` to do so.
```
nlarew, line 615: This shouldn't link to `$geoWithin` in the manual - that's a MongoDB geospatial operator. What we want instead is to link to the (not yet existing) Geospatial operators section of the RQL page.

source/schemas/types.txt
```diff
@@ -509,3 +509,123 @@ with the ``title`` field set to the embedded object's type name:
...
+Although GeoPoints are the only stored type you can query, your query will often
```
nlarew, line 616:
```suggestion
Although GeoPoints are the only stored type you can query, your query may
```

nitty but don't want to assume their workload

source/schemas/types.txt
```diff
@@ -509,3 +509,123 @@ with the ``title`` field set to the embedded object's type name:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+GeoPoints are the only form of geospatial data that Device Sync can 
+query. The Device Sync server validates every embedded GeoPoint property with 
+2 rules:
+
+- The "type" field can only be set to the string "Point" (case-sensitive), 
+- The "coordinates" field can only be set to an array of doubles. The first element 
+  in this array (longitude) must fall in the range of [-180, 180], 
+  and the second element (latitude) must fall in the range of [-90, 90].
```
nlarew, line 605: Aren't these just the definition of a GeoJSON point? Rather than restating the rules could we just refer to the description above it?

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+
+If a device writes an object that fails validation, it will be reverted with a 
+:ref:`compensating write error <flexible-sync-errors>`. If a document that doesn't 
+meet these requirements is written directly to the MongoDB collection, the server 
+marks that document as unsyncable.
+
+Querying
+````````
+You can query a GeoPoint using :ref:`Realm Query Language (RQL) <rql>`, 
+using ``$geoWithin`` to do so. Although GeoPoints are the only stored type you
```
jonatlee, line 617: [nit] `$geoWithin` is a MongoDB operator. The RQL equivalent does not contain a dollar sign

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+GeoPoints are the only form of geospatial data that Device Sync can 
+query. The Device Sync server validates every embedded GeoPoint property with 
+the rules outlined above.
+
+If a device writes an object that fails validation, it will be reverted with a 
+:ref:`compensating write error <flexible-sync-errors>`. If a document that doesn't 
+meet these requirements is written directly to the MongoDB collection, the server 
+marks that document as unsyncable.
```
jonatlee, line 612: [nit] Do you mind linking the unsyncable documents docs here? https://www.mongodb.com/docs/atlas/app-services/mongodb/internal-database-usage/#unsynced-documents

source/schemas/types.txt
```diff
@@ -509,3 +509,124 @@ with the ``title`` field set to the embedded object's type name:
...
+                "bsonType": "double"
+              }
+            }
+          }
+        }
+      }
+   }
+
+The following code block shows a document that follows the schema. It has a 
+single embedded GeoPoint object named "location". Both the containing object
```
jonatlee, line 586: [nit] I don't think this statement is accurate anymore (just the containing object has an extra property)

jonatlee (approved):
LGTM mod nits! Appreciate the hard work on this

nlarew (approved):
LGTM

PR #602, opened 2023-10-05
(DOCSP-33400): Edge Server: Add details for upgrading the Edge Server

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-33400

### Staged Changes

- [Edge Server - Configure](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/DOCSP-33400/edge-server/configure/#upgrade-the-edge-server-version)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/edge-server/configure.txt
+++ b/source/edge-server/configure.txt
@@ -392,16 +392,16 @@ server, as well as errors that may have occurred.
 Upgrade the Edge Server Version
 -------------------------------
 
-As Edge Server evolves, you may want to upgrade your Edge Server version
-to take advantage of new functionality.
+As the Edge Server evolves, you may want to upgrade your Edge Server 
+version to take advantage of new functionality.
 
 Prerequisites
 ~~~~~~~~~~~~~
 
 Upgrading the Edge Server version requires the host machine to have 
 `jq <https://jqlang.github.io/jq/>`__ installed. If you do not already
-have ``jq`` installed, running an install command tells you to install 
-``jq``.
+have ``jq`` installed, running an Edge Server upgrade command tells you 
+to install ``jq``.
 
 .. _edge-server-upgrade-commands:
 
--- a/source/edge-server/configure.txt
+++ b/source/edge-server/configure.txt
@@ -420,15 +420,6 @@ Upgrade Edge Server Procedure
 
 .. procedure::
 
-   .. step:: Stop the Edge Server
-
-      From the ``edge_server`` directory, run the command to stop the Edge 
-      Server:
-
-      .. code-block:: shell
-
-         make down
-
    .. step:: Run the upgrade command
 
       From the ``edge_server`` directory, run 
@@ -438,6 +429,9 @@ Upgrade Edge Server Procedure
       .. code-block:: shell
 
          make upgrade-minor
+
+      Depending on your hardware and network connection, the upgrade 
+      process may take seconds to minutes.
 
    .. step:: Start the Edge Server
 
@@ -451,14 +445,8 @@ Upgrade Edge Server Procedure
    .. step:: (Optional) Reset the Edge Server
 
       If you encounter unexpected behavior after upgrading the Edge Server,
-      you may want to reset it and start it again. Stop the Edge Server
-      with:
-
-      .. code-block::
-
-         make down
-
-      Then, run the following command to reset and restart it:
+      you may want to reset it and start it again. Run the following 
+      command:
 
       .. code-block::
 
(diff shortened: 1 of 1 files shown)


source/edge-server/configure.txt
```diff
@@ -386,3 +386,80 @@ server, as well as errors that may have occurred.
             "Item": "truepredicate"
          }
       }
+
+.. _upgrade-edge-server-version:
+
+Upgrade the Edge Server Version
+-------------------------------
+
+As Edge Server evolves, you may want to upgrade your Edge Server version
```
rukumar333, line 395: Should this say `As the Edge Server evolves`?

source/edge-server/configure.txt
```diff
@@ -386,3 +386,80 @@ server, as well as errors that may have occurred.
...
+to take advantage of new functionality.
+
+Prerequisites
+~~~~~~~~~~~~~
+
+Upgrading the Edge Server version requires the host machine to have 
+`jq <https://jqlang.github.io/jq/>`__ installed. If you do not already
+have ``jq`` installed, running an install command tells you to install
```
rukumar333, line 403:
```suggestion
have ``jq`` installed, running an Edge Server upgrade tells you to install 
```

source/edge-server/configure.txt
```diff
@@ -386,3 +386,80 @@ server, as well as errors that may have occurred.
...
+.. _edge-server-upgrade-commands:
+
+Upgrade Edge Server Commands
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can upgrade the Edge Server version using one of three commands:
+
+- ``make upgrade-major``: Upgrade to the latest major version of Edge Server.
+- ``make upgrade-minor``: Upgrade to the latest minor version of Edge Server.
+- ``make upgrade-force``: Upgrade to the latest version of Edge Server,
```
rukumar333, line 415: @sudssm Do we want customers to use the force feature on their own? I assumed this "ok" to include until we officially release. What do you think?

source/edge-server/configure.txt
```diff
@@ -386,3 +386,80 @@ server, as well as errors that may have occurred.
...
+
+      .. code-block::
+
+         make down
+
+      Then, run the following command to reset and restart it:
+
+      .. code-block::
+
+         make clean up
```
rukumar333, line 465: Technically, customers don't need to run `make down` here since `make clean` will already stop the server. @sudssm Want to confirm once with you that it is reasonable to remove the first part of this.

source/edge-server/configure.txt
dacharyc: Related to this, I saw that major version upgrades aren't supported yet. I started to add a note about that here, but then decided to leave it out. Since that message is available when you try to run the command, I thought I could leave it to the script to provide that info. That means we won't miss updating it in the docs when you do add that support. But I'm open to adding that detail here if you think it's important to communicate in the docs (or if we want to say anything about why).

source/edge-server/configure.txt
```diff
@@ -386,3 +386,80 @@ server, as well as errors that may have occurred.
...
+  regardless of major or minor version status.
+
+Upgrade Edge Server Procedure
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. procedure::
+
+   .. step:: Stop the Edge Server
+
+      From the ``edge_server`` directory, run the command to stop the Edge
```
nlarew, line 425:
```suggestion
      From the ``edge_server`` directory, run the following command to stop the Edge 
```

source/edge-server/configure.txt
```diff
@@ -386,3 +386,80 @@ server, as well as errors that may have occurred.
...
+
+      .. code-block:: shell
+
+         make down
+
+   .. step:: Run the upgrade command
+
+      From the ``edge_server`` directory, run 
+      :ref:`the relevant upgrade command <edge-server-upgrade-commands>`
+      to upgrade to your preferred version of the Edge Server:
```
nlarew, line 436: Do we have any indications of how long the upgrade might take? Could be a nice bit of info to add here.

nlarew (approved):
LGTM w/ a couple nits

source/edge-server/configure.txt
sudssm: I dont' have a strong preferences either way, so we can leave this as-is

source/edge-server/configure.txt
sudssm: Stopping is not strictly required, the upgrade script will stop it for you. So you can omit this step

source/edge-server/configure.txt
sudssm: seconds to minutes

source/edge-server/configure.txt
sudssm: agree, we can drop the `down`

PR #605, opened 2023-10-10
Add adminListServices endpoint

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-33421

### Staged Changes

https://preview-mongodbmongocaleb.gatsbyjs.io/atlas-app-services/yaml_test/admin/api/v3/#tag/services/operation/adminListServices

## Note
The scope of this ticket expanded to including a new endpoint and updating the response objects for several endpoints.

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -2944,12 +2944,12 @@ paths:
                     description: The group ID sent in the request.
                   start:
                     type: string
-                    example: 2022-10-27T14:01:01Z
+                    example: "2022-10-27T14:01:01Z"
                     description: The date and time (ISO-8601 format) of the earliest requested
                       measurement.
                   end:
                     type: string
-                    example: 2022-12-27T14:01:01Z
+                    example: "2022-12-27T14:01:01Z"
                     description: The date and time (ISO-8601 format) of the latest requested
                       measurement.
                   granularity:
@@ -8283,13 +8283,13 @@ components:
                 The HTTP method used by the incoming request.
             started:
               type: string
-              example: 2022-12-08T18:24:48.409Z
+              example: "2022-12-08T18:24:48.409Z"
               description: >
                 An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp for
                 the start of the logged operation.
             completed:
               type: string
-              example: 2022-12-08T18:24:48.41Z
+              example: "2022-12-08T18:24:48.41Z"
               description: >
                 An [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp for
                 the completion of the logged operation.
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -886,7 +886,7 @@ paths:
           application/json:
             schema:
               oneOf:
-                - $ref: "#/components/schemas/ServiceResponse"
+                - $ref: "#/components/schemas/CreateServiceRequest"
                 - $ref: "#/components/schemas/AtlasFederatedInstance"
                 - $ref: "#/components/schemas/ThirdPartyService"
       responses:
@@ -5570,8 +5570,7 @@ components:
               example: atlas
               description: The type of service.
             flexible_sync:
-              type: object
-              properties: "null"
+              $ref: "#/components/schemas/FlexibleSync"
             groupName:
               type: string
               example: Project 0
@@ -5600,11 +5599,13 @@ components:
           type: object
           description: "[Deprecated] The configuration for a third-party service."
     ServiceResponse:
-      _id: string;
-      name: string;
-      type: string;
-      version: number;
-      last_modified: number;
+      type: object
+      properties:
+        _id: string;
+        name: string;
+        type: string;
+        version: number;
+        last_modified: number;
     ReadPreferencesTagSet:
       type: string
       enum:
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -846,10 +846,8 @@ paths:
             application/json:
               schema:
                 items:
-                  anyOf:
-                    - $ref: "#/components/schemas/ServiceResponse"
-                    - $ref: "#/components/schemas/AtlasFederatedInstance"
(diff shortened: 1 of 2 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -847,48 +847,70 @@ paths:
               schema:
                 items:
                   anyOf:
-                    - $ref: "#/components/schemas/AtlasCluster"
+                    - $ref: "#/components/schemas/ServiceResponse"
```
nlarew, line 850: I think we should remove the `anyOf` here and only use `ServiceResponse`

Screenshot 2023-10-12 at 11 39 38 AM

source/openapi-admin-v3.yaml
```diff
@@ -5525,7 +5547,71 @@ components:
...
+        name: string;
+        type: string;
+        version: number;
+        last_modified: number;
+    ReadPreferencesTagSet:
+      type: string
+      enum:
+        - server1
+        - server2
+    CreateServiceRequest:
```
nlarew, line 5614: This seems to be specifically for data sources - can we use a less generic name? Also maybe an opportunity to reuse the `AtlasCluster` schema?

source/openapi-admin-v3.yaml
```diff
@@ -5539,35 +5625,49 @@ components:
           enum:
             - mongodb-atlas
         config:
+          config: "null"
```
nlarew, line 5628: Typo? I'm not familiar with the `config` field and not sure what "null" would mean here

source/openapi-admin-v3.yaml
```diff
@@ -5600,7 +5700,7 @@ components:
         - type
         - config
       oneOf:
-        - $ref: "#/components/schemas/AtlasCluster"
+        - $ref: "#/components/schemas/AppConstructor"
```
nlarew, line 5703: This one should still be `AtlasCluster`.

`AppConstructor` reference `DataSource` so this is circular as-is.

source/openapi-admin-v3.yaml
```diff
@@ -5525,7 +5547,71 @@ components:
         error_code:
           type: string
           description: The error type.
-    AtlasCluster:
+    AtlasConfigResponse:
```
nlarew, line 5550: Can we still use `AtlasCluster` here? It's a bit more descriptive (this doesn't represent Atlas Federated Instances for example).

source/openapi-admin-v3.yaml
```diff
@@ -5525,7 +5547,71 @@ components:
...
+          properties:
+            clusterId:
+              type: number
+              example: 05e821b3975d271289f372e3a
+              description: The service ID.
+            clusterName:
+              type: string
+              example: Cluster0
+              description: |-
+                The service name used to refer to the data source within
```
nlarew, line 5564: Cluster Name !== Service Name (though eventually they say they will be). This is the actual name of the cluster in Atlas, not e.g. "mongodb-atlas".

source/openapi-admin-v3.yaml
```diff
@@ -5525,7 +5547,71 @@ components:
...
+                this Atlas app. The name may be at most 64 characters
+                long and must only contain ASCII letters, numbers,
+                underscores, and hyphens.
+            clusterType:
+              type: string
+              example: atlas
+              description: The type of service.
```
nlarew, line 5571:
```suggestion
            clusterType:
              type: string
              enum:
                - atlas
                - serverless
              example: atlas
              description: The type of service.
```

There's only two options here so we can use an enum

nlarew (changes requested):

source/openapi-admin-v3.yaml
MongoCaleb: Forgot to remove that -- it was a placeholder until I figured out the details. :( Good catch!

source/openapi-admin-v3.yaml
MongoCaleb: I'd love to know why.

source/openapi-admin-v3.yaml
```diff
@@ -850,45 +869,46 @@ paths:
                     - $ref: "#/components/schemas/AtlasCluster"
                     - $ref: "#/components/schemas/AtlasFederatedInstance"
                     - $ref: "#/components/schemas/ThirdPartyService"
-    post:
+    patch:
```
nlarew, line 872: We lost the real `adminCreateService` endpoint somewhere here - there should be an endpoint for `POST /groups/{groupId}/apps/{appId}/services/config`.

Let's re-add that endpoint and then update this one to be `adminUpdateServiceConfig` or similar with an appropriate `summary`.

MongoCaleb: gah! Moving things around so much has led to quite the mess.

source/openapi-admin-v3.yaml
```diff
@@ -882,13 +884,70 @@ paths:
...
+          description: Successfully created.
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: "#/components/schemas/ServiceResponse"
+                  - $ref: "#/components/schemas/AtlasFederatedInstance"
+                  - $ref: "#/components/schemas/ThirdPartyService"
+        "409":
+          description: There is already a service with the given `name`.
```
nlarew, line 939:
```suggestion
        "204":
          description: Successfully updated.
```

Looks like this still has the createService responses. Just checked and the update config endpoint returns 204 No Content

Screenshot 2023-10-13 at 5 40 47 PM

source/openapi-admin-v3.yaml
```diff
@@ -898,7 +957,7 @@ paths:
             application/json:
               schema:
                 oneOf:
-                  - $ref: "#/components/schemas/AtlasCluster"
+                  - $ref: "#/components/schemas/AppConstructor"
```
nlarew, line 960:
```suggestion
                  - $ref: "#/components/schemas/AtlasCluster"
```

source/openapi-admin-v3.yaml
nlarew: > I'd love to know why.

Per that screenshot, the endpoint only returns `ServiceResponse` objects

nlarew (approved):
LGTM

PR #611, opened 2023-10-16
New Database Trigger Options

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-33588

### Staged Changes

https://preview-mongodbmongocaleb.gatsbyjs.io/atlas-app-services/DOCSP-33588/triggers/database-triggers/#std-label-trigger-recursion


source/triggers/database-triggers.txt
```diff
@@ -137,81 +151,130 @@ Database Triggers have the following configuration options:
...
+     - Options
 
-   * - :guilabel:`Document Preimage`
+   * - :guilabel:`Collection`
+     - |
+       - :guilabel:`Cluster Name`. The name of the MongoDB cluster that the 
+           Trigger is associated with.
+      
+       - :guilabel:`Database Name`. The MongoDB database that contains the watched
+         collection. Optional if the :guilabel:`Collection Name` is specified.
```
nathantfrank, line 187:
```suggestion
         collection. Required if the :guilabel:`Collection Name` is specified.
```

Maybe we should add some copy about how a Deployment trigger is when `database and collection` are not specified. Database trigger is when `collection` is not specified, and Collection Trigger when both are specified.

source/triggers/database-triggers.txt
```diff
@@ -137,81 +151,130 @@ Database Triggers have the following configuration options:
...
-     - If enabled, change events include a copy of the modified document
-       from immediately *before* the change was applied in the
-       ``fullDocumentBeforeChange`` field. All change events except for
-       **Insert** events include the document preimage. Document preimages are 
-       supported on non-sharded Atlas clusters running MongoDB 4.4+, and 
-       on sharded Atlas clusters running MongoDB 5.3 and later.
+       - :guilabel:`Collection Name`. The MongoDB collection to watch.
 
-       .. tip:: Performance Optimization
+       - :guilabel:`Operation Type`. The :ref:`database operation types
```
nathantfrank, line 191: [Q] should this be `database`?

source/triggers/database-triggers.txt
MongoCaleb: @nathantfrank does this mean that even if the user selects the Collection or Database source type, it will actually be a Deployment type if they leave both `database` and `collection` blank?

source/triggers/database-triggers.txt
MongoCaleb: @nathantfrank Sorry, not sure what you're asking here.

source/triggers/database-triggers.txt
nathantfrank: Ahhh Sorry I misunderstood what this section was doing.
My current thought is we should just remove that sentence.
 ```suggestion
         collection.
```

File: source/triggers/database-triggers.txt (hunk 2 above)
Comment (nathantfrank) [reply source/triggers/database-triggers.txt:191]: Ahhh, I think I'm confusing myself with existing documentation details. I was thinking this should be like `collection operation types` and link to something different.

File: source/triggers/database-triggers.txt [hunk 3]
```diff
@@ -137,81 +151,144 @@ Database Triggers have the following configuration options:
...
-       .. note::
+       - :guilabel:`Operation Type`. The collection operation types 
+         that cause the Trigger to fire. Select the operation types you want the 
+         trigger to respond to. Options include:
+       
+         - Create Collection
+         - Modify Collection
+         - Rename Collection
+         - Shard Collection
+         - Drop Collection
```
nathantfrank, line 200: As mentioned in our chat, these are actually Database+ change event types.

source/triggers/database-triggers.txt
```diff
@@ -137,81 +151,141 @@ Database Triggers have the following configuration options:
...
-       the ``fullDocument`` field.
+       - :guilabel:`Operation Type`. The collection operation types 
+         that cause the Trigger to fire. Select the operation types you want the 
+         trigger to respond to. Options include:
+       
+         - Create Document
+         - Modify Document
+         - Rename Document
+         - Shard Document
+         - Drop Document
```
nathantfrank, line 201: This should be Insert, Update, Replace, and Delete

source/triggers/database-triggers.txt
```diff
@@ -137,81 +151,141 @@ Database Triggers have the following configuration options:
...
+       - :guilabel:`Operation Type`. The :ref:`database operation types 
+         <database-event-operation-types>` that cause the Trigger to fire. 
+         Select the operation types you want the  trigger to respond to. 
+         Options include:
+       
+         - Create Collection
+         - Modify Collection
+         - Rename Collection
+         - Shard Collection
+         - Drop Collection
```
nathantfrank, line 243: Missing the 2 new sharding types

nathantfrank (changes requested):
Also just talked to Product about this recent change. Deployment level will not be supported if the cluster being watched is a shared tier. Not exactly sure where we want to add this in the docs.

MongoCaleb:
@nathantfrank : WRT your comment about Deployment only being supported on non-shared tiers, what happens in the UI? Is that box just diabled? What happens if I choose, say, Collection, and then leave Collection Name and DB Name blank (which _should_ force Deploy level)? Thanks!

source/triggers/database-triggers.txt
MongoCaleb: Doh!

nathantfrank:
Screen Shot 2023-10-18 at 10 58 53 AM
Screen Shot 2023-10-18 at 10 58 34 AM

> WRT your comment about Deployment only being supported on non-shared tiers, what happens in the UI? Is that box just diabled? 

Images taken from local branch, not yet merged. So it will be disabled. If they select Deployment and then select a shared tier we will move them automatically to Database level.

> What happens if I choose, say, Collection, and then leave Collection Name and DB Name blank (which should force Deploy level)?

If they try to save as described above we have added validation on the backend to respond with: `deployment changestreams are not supported on shared tier clusters`. This is the suspended trigger error the customer will also receive if they managed to get a deployment trigger with a shared tier cluster (somehow bypassing validation) on their app.

source/triggers/database-triggers.txt
```diff
@@ -137,81 +151,148 @@ Database Triggers have the following configuration options:
...
+       - :guilabel:`Cluster Name`. The name of the MongoDB cluster that the 
+         Trigger is associated with.
+       
+       - :guilabel:`Database Name`. The MongoDB database to watch. Optional. 
+         If you leave this option blank, the Source Type changes to "Deployment," 
+         unless you are on a shared tier, in which case App Services will not 
+         let you save the trigger.
+
+       - :guilabel:`Operation Type`. The :ref:`database operation types 
+         <database-event-operation-types>` that cause the Trigger to fire.
```
nathantfrank, line 240: [Q] Only this Database row has a link, do we want to have a link for `Collection` and `Deployment` as well?

nathantfrank (commented):
All most there from my perspective just 2 questions.

[Q] `Use Match Expressions to Limit Trigger Invocations` only lists the document change event types, do we want to include links for the others?

source/triggers/database-triggers.txt
```diff
@@ -53,6 +43,27 @@ expressions to limit the data included in each event.
...
+.. _trigger-recursion:
+
+.. warning::
+
+   It is possible to configure triggers that cause other triggers to fire on the 
+   same collection, database, or cluster, resulting in recursion. Examples 
+   include a database trigger writing to a collection within the same 
+   database, or a cluster-level logger writing database logs to another database 
+   in the same cluster.
+
```
nathantfrank, line 66: I think we may want to explicitly call out log forwarders as well here. Since, they are a first class resource that our users may also be using that could cause this issue.

nathantfrank (changes requested):
Are we able to time the release or is once this is merged it is live in the documentation?

source/triggers/database-triggers.txt
MongoCaleb: Isn't that just a specific case of a process writing to a monitored database and/or collection?

MongoCaleb:
@nathantfrank we can delay deployment. When is the anticipated release?

source/triggers/database-triggers.txt
nathantfrank: You are correct, but just wanted to make customer's slightly more conscious if they didn't make that connection.

nathantfrank:
> @nathantfrank we can delay deployment. When is the anticipated release?

We are targeting the week after skunkworks, so likely around Nov 1st would be ideal.

nlarew (approved):

source/triggers/database-triggers.txt
nathantfrank: Talked offline in the project channel with team and Laura, we would like to go with.

```suggestion
.. warning::

   In deployment and database level triggers, it is possible to configure triggers 
   in a way that causes other triggers to fire, thereby resulting in recursion. 
   Examples include a database-level trigger writing to a collection within the 
   same database, or a cluster-level logger or log forwarder writing logs to 
   another database in the same cluster.
   
```

nathantfrank: @MongoCaleb Could we get an update on this since it should be going in tomorrow.

source/triggers/database-triggers.txt
```diff
@@ -272,23 +392,52 @@ following operation types:
...
 
-   * - **Delete**
+   * - **Delete Document** (All trigger types)
      - Represents a document deleted from the collection.
 
-   * - **Replace**
+   * - **Replace Document** (All trigger types)
      - Represents a new document that replaced a document in the collection.
 
+   * - **Create_Collection** (Database and Deployment trigger types only)
```
nathantfrank, line 412: [Q] Any particular reason we are underscoring here?

nathantfrank (changes requested):
1 Q/potential change, then looks good to me when that is addressed.

nathantfrank (approved):

PR #619, opened 2023-11-01
(DOCSP-33503) Unhide Find Your App ID CLI procedure

nlarew:
### Jira

- (DOCSP-33503) Unhide Find Your App ID CLI procedure

### Staged Changes

- [Apps > Metadata](https://docs-atlas-staging.mongodb.com/atlas-app-services/docsworker-xlarge/unhide/apps/metadata/)

### DIFF
--- a/source/apps/metadata.txt
+++ b/source/apps/metadata.txt
@@ -192,7 +192,7 @@ or Atlas CLI to find it programmatically.
                :alt: Expanded ellipses icon showing the Actions menu and the copy
                :width: 400px
 
-   .. tab:: CLI
+   .. tab:: Atlas CLI
       :tabid: cli
 
       You can list all Atlas projects that you have access to with the
--- a/source/apps/metadata.txt
+++ b/source/apps/metadata.txt
@@ -106,35 +106,39 @@ or App Services CLI to find it programmatically.
          Make sure to replace ``<access_token>`` and ``{groupId}`` with
          your own values.
 
-..   .. tab::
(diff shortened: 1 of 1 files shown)


source/apps/metadata.txt
```diff
@@ -188,7 +192,7 @@ or Atlas CLI to find it programmatically.
                :alt: Expanded ellipses icon showing the Actions menu and the copy
                :width: 400px
 
-   .. tab:: Atlas CLI
+   .. tab:: CLI
```
dacharyc, line 195: Hmm. I noticed the tab title on the page for the procedure we're un-hiding above is "App Services CLI" and the tab title of this procedure is just "CLI." It looks like we're using `.. tabs-realm-admin-interfaces::` above, and for this procedure, we're just using `.. tabs::` in ln 155. Should we use the same tab group as above for consistency?

dacharyc (approved):
Overall LGTM - I just noticed we're using different tab groups here, which means inconsistent tab naming and it also means that when a user clicks the CLI tab in the first procedure, we're not showing the CLI tab in the second procedure. Maybe we should switch the second procedure to use the same tab group?

source/apps/metadata.txt
nlarew: This procedure can't use the realm-admin-interfaces tabset because it's using all the Atlas stuff. It'd be cool if we could link the App Services CLI and Atlas CLI tabs together but alas they are not currently connected.

PR #623, opened 2023-11-02
(DOCSP-33859): Edge Server: Add required networking details

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-33859

### Staged Changes

- [Configure Edge Server](https://preview-mongodbdacharyc.gatsbyjs.io/atlas-app-services/DOCSP-33859/edge-server/configure/#required-network-access)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/edge-server/configure.txt
+++ b/source/edge-server/configure.txt
@@ -113,8 +113,19 @@ server, or hardware that you provision on-site.
 Required Network Access
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-The host that runs the Edge Server must allow access to port 80 via HTTP 
-to the general internet. 
+To accept connections from sync clients or connections via the wireprotocol, 
+the host that runs the Edge Server must allow access to incoming connections.
+
+While Edge Server is in Preview, you can stand up a demo using EC2 or another 
+cloud provider by opening port 80 via HTTP to the general internet. 
+In production, we recommend opening port 443 with SSL.
+
+If you run Edge Server on your own infrastructure, you don't need to open
+the instance to the general internet. You only need to ensure your client 
+devices can access it.
+
+Edge Server supports TLS. You can find configuration details in the 
+"Complete the Edge Server Configuration Details" section.
 
 .. _edge-server-os-support:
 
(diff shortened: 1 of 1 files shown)


source/edge-server/configure.txt
```diff
@@ -113,8 +113,19 @@ server, or hardware that you provision on-site.
...
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-The host that runs the Edge Server must allow access to port 80 via HTTP 
-to the general internet. 
+To accept connections from sync clients or connections via the wireprotocol, 
+the host that runs the Edge Server must allow access to incoming connections.
+
+While Edge Server is in Preview, you can stand up a demo using EC2 or another 
+cloud provider by opening port 80 via HTTP to the general internet. 
+In production, we recommend opening port 443 with SSL.
```
krollins-mdb, line 121: There's an implicit suggestion that Preview === dev environment. Do we know if this is true? Beyond that, is using HTTP in a dev environment going away after Edge Server moves out of Preview? If not, we may want to refer to "using Edge Server in a development environment" instead of "While Edge Server is in Preview". Also, maybe we should address the "general internet vs owned infrastructure" a bit earlier?

Here's something to show what I'm getting at:

```suggestion
If you need to open Edge Server to the genera internet, you should open different ports with different methods appropriate for your environment.

- Development. Use port 80 and HTTP.
- Production. Use port 443 and SSL.
```

source/edge-server/configure.txt
```diff
@@ -113,8 +113,19 @@ server, or hardware that you provision on-site.
...
+
+If you run Edge Server on your own infrastructure, you don't need to open
+the instance to the general internet. You only need to ensure your client 
+devices can access it.
+
+Edge Server supports TLS. You can find configuration details in the 
+"Complete the Edge Server Configuration Details" section.
```
krollins-mdb, line 128: Are using the "see <section-name> on this page" construction? Or was that only if we're including links to other sections on the same page?

krollins-mdb (approved):
Nice work, Dachary! No blocking comments, but I do have some questions and suggestions.

source/edge-server/configure.txt
dacharyc: Good call - reworded and reorganized this section a little.

PR #629, opened 2023-11-16
DOCSP-34316 Update Role Example Descriptions

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-34316

### Staged Changes

- [Data Access Role Examples](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-34316/rules/examples/#crud-permissions)

### Reminder Checklist

### Review Guidelines


source/images/role-template-cannot-insert-new-docs.png
dacharyc, line 1: This image doesn't quite look like the one it's replacing. In the current docs, read/write with no insert shows that `Delete` is checked but `Insert` is unchecked: https://www.mongodb.com/docs/atlas/app-services/rules/examples/

In the staging site, it's just read/write all fields - `Delete` isn't checked: https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-34316/rules/examples/#the-role-can-read---write-all-fields-but-cannot-insert-new-documents

Should this match what's in the existing docs? i.e. if you set up this screenshot in the Rules UI, does the JSON for this match what's in the docs?

```json
{
  "name": "<Role Name>",
  "apply_when": <JSON Expression>,
  "insert": false,
  "delete": <boolean>,
  "write": true,
}
```

dacharyc (approved):
Overall, this LGTM! But I do have a question about one of the images where it doesn't match what's in the current docs. If that is correct, then I think this is good to go - otherwise, we might need to update that one screenshot.

source/images/role-template-cannot-insert-new-docs.png
lindseymoore: As delete is not mentioned in this example, it is fine to leave delete unchecked. The JSON is correct for the image in the PR.

PR #650, opened 2023-12-14
DOCSP-34996 External release notes for v20231213

lindseymoore:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-34996

- [Release Notes > 13 December 2023 Release](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-34996/release-notes/backend/#13-december-2023-release)

### Review Guidelines

### DIFF
--- a/source/release-notes/backend.txt
+++ b/source/release-notes/backend.txt
@@ -18,29 +18,19 @@ Atlas App Services Changelog
 13 December 2023 Release
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-- Added support for passing in user's new metadata from a request using JWT authentication to the function context.
-
-- Apps, users, and refresh token are now pruned at the 30 day mark.
-
-- Fixed webpack production yarn build warnings, such as "asset size limit" and "entrypoint size warnings", by reducing bundle size. 
+- Requests authenticated through the Custom JWT provider now update the user's profile data (e.g. ``context.user.data``) based on information from the JWT.
 
 - Updated App Services schema UI to display the difference between a 64-bit ``long`` and 32-bit ``integer``.
 
 - Disallowed creating schemas with invalid database or collection names, such as names including spaces.
 
-- Left default for ``allowDiskUse`` as true unless user specifies it as false. 
+- MongoDB aggregation operations no longer specify a default value for ``allowDiskUse``.
 
 - Fixed bug where a read document filter set to false is not unioned with a write filter.
 
-- Created a private endpoint for enabling tiered sync.
-
 - Added ``integer`` as a supported type for indexed queryable fields.
 
-- Updated the merge window filtering to use indexed queryable fields if possible.
-
-- Distributed translators evenly within the sync pool.
-
-- If one of several sync translators has failed, the failed subscription banner will now show up on the sync page.
+- Device Sync permissions now properly union a read filter set to ``false`` with its corresponding write filter.
 
 .. _backend_20231129:
 
--- a/source/release-notes/backend.txt
+++ b/source/release-notes/backend.txt
@@ -24,7 +24,7 @@ Atlas App Services Changelog
 
 - Fixed webpack production yarn build warnings, such as "asset size limit" and "entrypoint size warnings", by reducing bundle size. 
 
-- Updated App Services schema UI to display the difference between 64-bit long and 32-bit integer data types.
+- Updated App Services schema UI to display the difference between a 64-bit ``long`` and 32-bit ``integer``.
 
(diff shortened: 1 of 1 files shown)


source/release-notes/backend.txt
```diff
@@ -13,6 +13,35 @@ Atlas App Services Changelog
 2023 Releases
 -------------
 
+.. _backend_20231213:
+
+13 December 2023 Release
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Added support for passing in user's new metadata from a request using JWT authentication to the function context.
```
nlarew, line 21:
```suggestion
- Requests authenticated through the Custom JWT provider now update the user's profile data (e.g. ``context.user.data``) based on the JWT.
```

I think we can focus this less on the implementation and more on its impact to the reader. Also this is kind of a bugfix improvement so maybe we can frame it more like that.

source/release-notes/backend.txt
```diff
@@ -13,6 +13,35 @@ Atlas App Services Changelog
...
+
+- Apps, users, and refresh token are now pruned at the 30 day mark.
+
+- Fixed webpack production yarn build warnings, such as "asset size limit" and "entrypoint size warnings", by reducing bundle size. 
+
+- Updated App Services schema UI to display the difference between a 64-bit ``long`` and 32-bit ``integer``.
+
+- Disallowed creating schemas with invalid database or collection names, such as names including spaces.
+
+- Left default for ``allowDiskUse`` as true unless user specifies it as false.
```
nlarew, line 31:
```suggestion
- MongoDB aggregation operations no longer specify a default value for ``allowDiskUse``.
```

If you check out [the code in the PR](https://github.com/10gen/baas/pull/12029/files#diff-fa3d6d98d704671b1cd8b1ed6516651388cd87db0dbb61dd8cb3eecbf36140abR238-R239) it's only setting the value if the user passes it.

source/release-notes/backend.txt
```diff
@@ -13,6 +13,35 @@ Atlas App Services Changelog
...
+
+- Updated App Services schema UI to display the difference between a 64-bit ``long`` and 32-bit ``integer``.
+
+- Disallowed creating schemas with invalid database or collection names, such as names including spaces.
+
+- Left default for ``allowDiskUse`` as true unless user specifies it as false. 
+
+- Fixed bug where a read document filter set to false is not unioned with a write filter.
+
+- Created a private endpoint for enabling tiered sync.
```
nlarew, line 35:
```suggestion
- Device Sync permissions now properly union a read filter set to ``false`` with its corresponding write filter.
```

1. For this one I think it's clearer to describe the new behavior than saying what it doesn't not do anymore

2. This is a private endpoint so best to leave it out of the docs

source/release-notes/backend.txt
```diff
@@ -13,6 +13,35 @@ Atlas App Services Changelog
...
+
+- Added ``integer`` as a supported type for indexed queryable fields.
+
+- Updated the merge window filtering to use indexed queryable fields if possible.
+
+- Distributed translators evenly within the sync pool.
+
+- If one of several sync translators has failed, the failed subscription banner will now show up on the sync page.
```
nlarew, line 43:
```suggestion
- Added ``integer`` as a supported type for indexed queryable fields.
```

These last three are pretty internal focused so let's remove them.

"merge window" is an implementation detail of operational transformation and is not a term we expose to the end users, so it doesn't really make much sense here. Also, the potential bug it fixes is very rare so I think we can just leave it out.

Similar idea with the other two - our k8s configuration and management of sync translators is something we abstract away from the end user.

nlarew (changes requested):

source/release-notes/backend.txt
```diff
@@ -13,6 +13,25 @@ Atlas App Services Changelog
...
+
+- Requests authenticated through the Custom JWT provider now update the user's profile data (e.g. ``context.user.data``) based on information from the JWT.
+
+- Updated App Services schema UI to display the difference between a 64-bit ``long`` and 32-bit ``integer``.
+
+- Disallowed creating schemas with invalid database or collection names, such as names including spaces.
+
+- MongoDB aggregation operations no longer specify a default value for ``allowDiskUse``.
+
+- Fixed bug where a read document filter set to false is not unioned with a write filter.
```
nlarew, line 29: This one is duped by the last bullet

nlarew (approved):
LGTM % one extra bullet point to remove

PR #663, opened 2024-01-29
DOCSP-35902 Security considerations for roles and query filtering

lindseymoore:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-35902

- [Role-based Permissions](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-35902/rules/roles/#define-roles---permissions): note at bottom of page about security considerations
- [Filter Incoming Queries](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-35902/rules/filters/#procedure): note at bottom of page about security considerations

### Reminder Checklist

  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

- Define Data Access Permissions
  - Role-based Permissions: Add note about security consideration for roles and filters for queries. 
  - Filter Incoming Queries: Add note about security consideration for roles and filters for queries.

### Review Guidelines

### DIFF
--- a/source/rules.txt
+++ b/source/rules.txt
@@ -11,7 +11,7 @@ Define Data Access Permissions
    
    Role-based Permissions </rules/roles>
    Rule Expressions </rules/expressions>
-   Filter Incoming Queries </rules/filters>
+   Filter Query Results </rules/filters>
    Device Sync-Compatible Permissions </rules/sync-compatibility>
    Configure Advanced Rules </mongodb/configure-advanced-rules>
    Data Access Role Examples </rules/examples>
--- a/source/rules.txt
+++ b/source/rules.txt
@@ -11,7 +11,7 @@ Define Data Access Permissions
    
    Role-based Permissions </rules/roles>
    Rule Expressions </rules/expressions>
-   Filter Query Results </rules/filters>
+   Filter Incoming Queries </rules/filters>
    Device Sync-Compatible Permissions </rules/sync-compatibility>
    Configure Advanced Rules </mongodb/configure-advanced-rules>
    Data Access Role Examples </rules/examples>
--- a/source/rules/roles.txt
+++ b/source/rules/roles.txt
@@ -11,6 +11,13 @@ Role-based Permissions
 
 .. default-domain:: mongodb
 
+.. meta::
+  :description: Learn how to define data access permissions using Role-based Permissions in App Services. 
+
+.. facet::
+  :name: genre 
+  :values: reference
+
 .. contents:: On this page
    :local:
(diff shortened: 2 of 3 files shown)


source/rules.txt
```diff
@@ -11,7 +11,7 @@ Define Data Access Permissions
    
    Role-based Permissions </rules/roles>
    Rule Expressions </rules/expressions>
-   Filter Incoming Queries </rules/filters>
+   Filter Query Results </rules/filters>
```
nlarew, line 14: Why the name change? These filters are the first step in the rules process but IMO the new name implies they're the final step.

nlarew: You're right that the end user will see the results of a Filter in the query result set that's returned to them.

In practice though the purpose of these is more to optimize document-level rule evaluation because it's a fairly expensive operation (i.e. it slows down your query) and it scales with the number of documents it needs to evaluate. So rather than evaluating rules for everything that matches an arbitrary user query, you can say "I _know_ that I can always filter out these additional documents under these conditions because they won't be returned to the user anyway."

That said, there may be a use case that's closer to "the end user will see the results of a Filter in the query result" for Device Sync rules that I haven't really thought about.

Either way, we may want to eventually reframe this around optimizing rule execution (or make a new section/page about it).

lindseymoore: Thanks! I see that the focus is more at the beginning of the process, on optimizing the query. I'll change the name back @nlarew

source/rules/roles.txt
```diff
@@ -511,3 +518,16 @@ deploying configuration files with {+cli+}:
...
+  While Role-based Permissions and :ref:`Filters <filters>` can hide specific 
+  documents and fields within a collection there is a potential that 
+  data can be exposed if the system allows arbitrary 
+  queries to access the collection. For example, queries or functions that 
+  raise errors depending on the values stored in a collection (such 
+  as division-by-zero errors) may reveal information about documents, even if 
+  a role or filter prevents the querying user from viewing 
+  documents directly. Users may also make inferences about the underlying data 
+  in other ways (such as by measuring query execution time, which can be affected 
+  by the data's distribution).
```
nlarew, line 533: In terms of what's in the ticket this covers everything. However, I think it may be a bit frustrating for devs to read this and not really have anything actionable.

Maybe we can add a final sentence that's basically like "be aware that this is possible and audit your data access patterns if it's a concern for you"

nlarew (approved):
LGTM with one possible addition, up to you

PR #675, opened 2024-02-14
DOCSP-36193: Update Admin API Trigger endpoints with EventBridge error handler properties

cbullinger:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-36193

- [App Services Admin API](https://preview-mongodbcbullinger.gatsbyjs.io/atlas-app-services/docsp-36194-eventbridge-admin-api/admin/api/v3/#tag/triggers)  (reference page): Add new `error_handler` object to the Database Trigger type for all relevant Trigger endpoints.

### Reminder Checklist

- [ **N/A** ] Did you tag pages appropriately? 
  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

- **Reference**
  - App Services Admin API: Update page to support AWS EventBridge custom error handling feature by adding the new `error_handler` object property to Database type Trigger payloads in all relevant Trigger endpoints. 

### Review Guidelines

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -7121,7 +7121,7 @@ components:
                       description: If 'true', custom error handling is enabled for the AWS EventBridge trigger. 
                     function_id:
                       type: string
-                      description: The ID of the custom error handler invoked when the AWS EventBridge trigger fails and cannot be successfully retried.
+                      description: The ID of the function invoked when the AWS EventBridge trigger fails and cannot be successfully retried.
     AuthenticationTrigger:
       example:
         name: onNewApiKey
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -7111,7 +7111,10 @@ components:
                     serially.
             error_handler:
               type: object
-              description: An object that defines custom error handling for an AWS EventBridge trigger. Valid only for `AWS_EVENTBRIDGE` event processors. 
+              description: |-
+                An object that defines custom error handling for an AWS 
+                EventBridge trigger. Valid only for `"DATABASE"` type triggers 
+                with `"AWS_EVENTBRIDGE"` event processors. 
               properties:
                 config:
                   type: object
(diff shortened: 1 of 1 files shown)


lindseymoore (approved):
Thanks, LGTM!

source/openapi-admin-v3.yaml
```diff
@@ -7109,6 +7109,19 @@ components:
                     can process events in parallel. If `false`, event
                     ordering is enabled and the Trigger executes events
                     serially.
+            error_handler:
+              type: object
+              description: An object that defines custom error handling for an AWS EventBridge trigger. Valid only for `AWS_EVENTBRIDGE` event processors.
```
tcannon91, line 7114: Is it worth adding a note (or adjusting the current note) that this is only valid for Database triggers as well?

Currently we support AWS EventBridge processors on all of our trigger types (Database, scheduled, auth), but that isn't out long term plan -- we plan to only support eventbridge for Database triggers eventually. So with that future in mind, we only implemented support for Custom Error Handlers with Database triggers.

source/openapi-admin-v3.yaml
```diff
@@ -7109,6 +7109,19 @@ components:
...
+              properties:
+                config:
+                  type: object
+                  properties:
+                    enabled:
+                      type: boolean
+                      description: If 'true', custom error handling is enabled for the AWS EventBridge trigger. 
+                    function_id:
+                      type: string
+                      description: The ID of the custom error handler invoked when the AWS EventBridge trigger fails and cannot be successfully retried.
```
tcannon91, line 7124: Just a thought -- custom error handlers are just functions at the end of the day. Would it be more clear to reword this to something closer to this:
```suggestion
                      description: The ID of the function invoked when the AWS EventBridge trigger fails and cannot be successfully retried.
```
If it isn't confusing due to the structure of the object or if we are trying to communicate something by using `custom error handler` then I am on board :) My suggestion is to hopefully alleviate any confusion that they need to look for custom error handler in the UI/app config rather than just any function they have created.

source/openapi-admin-v3.yaml
cbullinger: This object is only defined in the database type schema, but it wouldn't hurt to reiterate it here, as well. I'll update

Also, I meant to ask: is this programmatically enforced as invalid for the other trigger types?

source/openapi-admin-v3.yaml
cbullinger: No, but I was trying to be consistent with the naming. Agree it's potentially confusing here.

source/openapi-admin-v3.yaml
tcannon91: Yes, it is enforced by the api

PR #677, opened 2024-02-15
Update OpenAPI adminLogin endpoint

philmodin:
## Pull Request Info

Current OpenAPI spec doesn't account for `mongodb-cloud` as it should according to documentation in [Get an Admin API Session Access Token](https://www.mongodb.com/docs/atlas/app-services/admin/api/v3/#section/Get-an-Admin-API-Session-Access-Token)

The example specifies `mongodb-cloud` as the provider but it's not included in `ProviderType` in the spec file.

Example: https:// services.cloud.mongodb.com/api/admin/v3.0/auth/providers/`mongodb-cloud`/login

~~This PR adds the necessary provider so generated clients can access the Atlas App Services Admin API.~~
This PR hardcodes the provider into the path and removes the need for parameters.

### Reminder Checklist

### Release Notes

- **Update adminLogin operation**
  - /auth/providers/mongodb-cloud/login

### Review Guidelines

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -173,7 +173,7 @@ paths:
                 items:
                   "$ref": "#/components/schemas/BaseAuthProvider"
       security: []
-  "/auth/providers/{provider}/login":
+  "/auth/providers/mongodb-cloud/login":
     post:
       tags:
         - admin
@@ -220,13 +220,6 @@ paths:
                   device_id:
                     type: string
       security: []
-    parameters:
-      - name: provider
-        description: The authentication provider to use.
-        in: path
-        required: true
-        schema:
-          "$ref": "#/components/schemas/ProviderType"
   "/auth/profile":
     get:
       tags:
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -6684,7 +6684,6 @@ components:
         - oauth2-facebook
         - custom-token
         - custom-function
-        - mongodb-cloud
     MessageState:
       type: string
       enum:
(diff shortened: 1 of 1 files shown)


nlarew:
Hey Philip thanks for the PR!

Based on our records it looks like you've already signed the [MongoDB Contributor Agreement](https://www.mongodb.com/legal/contributor-agreement) so we'd be happy to accept your contribution!

This PR unearthed a deeper issue with the API documentation that we should fix before merging. Specifically, the `adminLogin` endpoint currently _only_ supports login via the `mongodb-cloud` provider. The other providers are for an App's end users rather than the admin users developing the App.

Similarly, we use the `ProviderType` enum in other places where `mongodb-cloud` is not supported.

Would you mind updating your PR to:
- Not modify the `ProviderType` schema
- Replace the reference to `ProviderType` in the `provider` parameter of `adminLogin` with a schema that only allows the string `mongodb-cloud`, e.g.
  ```yaml
  type: string
  const: mongodb-cloud
  ```

philmodin:
@nlarew good point there. I can do that, but how about an approach that reduces configuration and knobs?
Instead of a new scheme with a single option, just bake the parameter into the path and remove the need for parameters.

Instead of `/auth/providers/{provider}/login`
Do `/auth/providers/mongodb-cloud/login`

nlarew:
@philmodin I checked with the engineering team and it sounds like we don't have any plans to support more admin auth providers so it should be fine to hardcode `mongodb-cloud` into the endpoint instead of the `{provider}` parameter. Good idea!

nlarew (approved):
Looks good to me! Thanks for the update 😊

PR #696, opened 2024-03-04
(DOCSP-36973): Add info about Device Sync websocket read limit

dacharyc:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-36973

- [Service Limitations](https://preview-mongodbdacharyc.gatsbyjs.io/atlas-app-services/DOCSP-36973/reference/service-limitations/#default-message-size-limit): Add information about the 20 MB Device Sync websocket read limit.

### Reminder Checklist

  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

- **Reference**
  - Service Limitations: Add information about the 20 MB Device Sync websocket read limit.

### Review Guidelines


source/reference/service-limitations.txt
```diff
@@ -189,6 +190,26 @@ Reading:
    When you intend to have many concurrent writers, have each writer
    work on a separate document. 
 
+.. _mongodb-service-limitations-sync-default-message-size-limit:
+
+Default Message Size Limit
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Device Sync has a default message size limit of {+sync-default-message-size-limit+}
+compressed for any messages sent from the client application. The message size
```
krollins-mdb, line 199: Extra "compressed" here?

```suggestion
for any messages sent from the client application. The message size
```

source/reference/service-limitations.txt
```diff
@@ -189,6 +190,26 @@ Reading:
...
+depends on the compressed size of changes sent from the client device. This is 
+not directly equivalent to the size of any document in the upload. In practice,
+however, the changeset size is generally on the same order of magnitude of
+the size of documents in the upload.
+
+This limit acts as a countermeasure against connections hogging too many 
+resources.
```
krollins-mdb, line 206: > ...hogging too many resources.

We shouldn't use a colloquial term like "hogging" for a couple of reasons:

- Translation. We need to keep in mind that the docs are now being translated into multiple languages. Plain English is better for this, partly due to the next point.
- Specificity. It's unclear what "hogging" means in this category. Using up compute, taking up bandwidth, etc.

Any chance we can be more specific with this sentence?

krollins-mdb (approved):
LGTM, but there's one comment I think you should consider and another that's just a nit.

source/reference/service-limitations.txt
dacharyc: I am not seeing an extra compressed here - is there something you're seeing that I'm missing?

dacharyc: The HELP ticket that is linked from this Jira ticket refers to both compressed and uncompressed size. Uncompressed size of the specific transaction was 83 MB, compressed size was 23 MB. The Sync Protocol docs go into detail about compression and delta sync: https://www.mongodb.com/docs/atlas/app-services/sync/details/protocol/#changeset

I think the point is to note that the size limit is based on the size of the changeset after the compression has been applied. So changesets that are greater than 20 MB on the client may be fine as long as the compressed size falls within the limit.

krollins-mdb: Gotcha. Maybe it makes sense to mention that the data is compressed automatically? Otherwise some users may think they need to manually compress the data.

dacharyc: It's a fair point that this isn't explained here and it would be easy to misinterpret. I added some additional verbiage and linked out to the relevant docs section for more info about the compression we use. Thanks for the feedback!

PR #697, opened 2024-03-04
(DOCSP-34560): Add information about duplicate Trigger events

dacharyc:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-34560

- [Triggers landing page](https://preview-mongodbdacharyc.gatsbyjs.io/atlas-app-services/DOCSP-34560/triggers/#diagnose-duplicate-events): Add a section about diagnosing duplicate events.

### Reminder Checklist

  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

- **Triggers**
  - Landing Page: Add a section about diagnosing duplicate events.

### Review Guidelines

### DIFF
--- a/source/triggers.txt
+++ b/source/triggers.txt
@@ -125,3 +125,25 @@ may not exceed the number of available change streams.
 
 :ref:`Learn more about the number of supported change streams for Atlas tiers.
 <mongodb-service-limitations-change-streams>`
+
+.. _trigger-diagnose-duplicate-events:
+
+Diagnose Duplicate Events
+-------------------------
+
+During normal Trigger operation, Triggers do not send duplicate events.
+However, when some failure or error conditions occur, Triggers may deliver 
+duplicate events. You may see a duplicate Trigger event when:
+
+- A server responsible for processing and tracking events experiences a 
+  failure. This failure prevents the server from recording its progress in a 
+  durable or long-term storage system, making it "forget" it has processed 
+  some of the latest events.
+- Using unordered processing where events 1-N are sent simultaneously. If 
+  event 9 fails and leads to Trigger suspension, events like event 10 might 
+  get processed again when the system resumes from event 9. This can lead to 
+  duplicates, as the system doesn't strictly follow the sequence of events and 
+  may reprocess already-handled events.
+
+If you notice duplicate Trigger events, check the :ref:`logs` for suspended 
+Triggers or server failures.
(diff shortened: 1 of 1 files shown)


source/triggers.txt
```diff
@@ -125,3 +128,25 @@ may not exceed the number of available change streams.
...
+
+During normal Trigger operation, Triggers do not send duplicate events.
+However, when some failure or error conditions occur, Triggers may deliver 
+duplicate events. You may see a duplicate Trigger event when:
+
+- A server responsible for processing and tracking events experiences a 
+  failure. This failure prevents the server from recording its progress in a 
+  durable or long-term storage system, making it "forget" it has processed 
+  some of the latest events.
+- Using unordered processing where events 1-N are sent simultaneously. If
```
lindseymoore, line 145:
```suggestion
- Using unordered processing where events 1 through N are sent simultaneously. If 
```
I read as "1 minus N" because I wasn't expecting it. The phrase could also be deleted, and it wouldn't make much of a difference, considering the variable N isn't used any where else.

source/triggers.txt
```diff
@@ -125,3 +128,25 @@ may not exceed the number of available change streams.
...
+However, when some failure or error conditions occur, Triggers may deliver 
+duplicate events. You may see a duplicate Trigger event when:
+
+- A server responsible for processing and tracking events experiences a 
+  failure. This failure prevents the server from recording its progress in a 
+  durable or long-term storage system, making it "forget" it has processed 
+  some of the latest events.
+- Using unordered processing where events 1-N are sent simultaneously. If 
+  event 9 fails and leads to Trigger suspension, events like event 10 might 
+  get processed again when the system resumes from event 9. This can lead to
```
lindseymoore, line 147:
```suggestion
  an event fails and leads to Trigger suspension, the proceeding events might 
  get processed again when the system resumes from the failed event. This can lead to 
```
Suggestion: Could get rid of event numbering to aid in clarity.

lindseymoore (approved):
LGTM! Can take another look if needed.

source/triggers.txt
dacharyc: Gotcha. The original example was "1-10" but I changed it to "N" because I didn't want to imply that 10 was some sort of limit on unordered processing. With this feedback, I see that the more concrete example is helpful, and I do think having the numbers explicit here clarifies the example, so I'll change it back to 10. 👍

source/triggers.txt
dacharyc: I think the numbering really helps to clarify the example, so I've just updated it to be more explicit based on your feedback above. Thank you!

PR #706, opened 2024-03-15
DOCSP-36931 Update Sync Config

lindseymoore:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-36931

- [Admin API](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-36931/admin/api/v3/#tag/sync/operation/adminGetFlexSync): updated Flexible Sync Response Object with missing fields
- [Configuration Files > Atlas Device Sync](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-36931/reference/config/sync/): add `asymmetric_tables` field
- [Pause or Terminate Sync](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-36931/sync/configure/pause-or-terminate-sync/): removed  `queryable_fields_version` from the page

### Reminder Checklist

- [ ] Did you tag pages appropriately?
  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

**Device Sync**
- Configure Sync/Pause or Terminate Sync: Remove `queryable_fields_version` from the example config files, as the option is not user-editable. 

**Reference**
- App Services Admin API: Update "Get Flexible Sync information" endpoint with missing response fields.
- Configuration Files/Atlas Device Sync: Add `asymmetric_tables` to configuration file for device sync.

### Review Guidelines


source/openapi-admin-v3.yaml
```diff
@@ -7221,6 +7242,18 @@ components:
           type: object
           description: |-
             **Deprecated**
+        indexed_queryable_fields_names:
```
tkaye407, line 7245: [nit] It seems like ideally we would put this below queryable_fields_names

source/openapi-admin-v3.yaml
```diff
@@ -7221,6 +7242,18 @@ components:
...
+          description: |-
+            A list of field names to use as the indexed queryable field. While this property is an array, Sync 
+            currently supports only one one indexed queryable field. Therefore, this array may contain at most 
+            one element.
+
+            The indexed queryable field must be present in the schema and be the same eligible field type in 
+            every collection you sync. The indexed queryable field name must also appear in queryable_fields_names 
+            since this is a global queryable field.
+          items:
+            type: string
```
tkaye407, line 7256: Seems like we are also missing asymmetric_tables as an option

source/openapi-admin-v3.yaml
tkaye407, line 7245: We are in the process of browning this field out for what its worth (permissions) and we will likely want to remove this in a few weeks

source/openapi-admin-v3.yaml
lindseymoore: Added!

source/openapi-admin-v3.yaml
lindseymoore: Thanks, can do! Is there a ticket related to `permissions` deprecation? So I can add to a ticket to remove the option

source/openapi-admin-v3.yaml
```diff
@@ -7191,10 +7191,31 @@ components:
...
       type: object
+      required: 
+        - state
+        - development_mode_enabled
       properties:
         state:
           type: string
-          description: The state of the sync service ("enabled", "disabled", or empty if Sync is not configured).
+          description: The state of the sync service (`"enabled"`, `"disabled"`, or empty if Sync is not configured).
+        development_mode_enabled:
```
tkaye407, line 7201: I need to confirm, but I think interestingly developer_mode_enabled and service_name are not included in this payload

tkaye407: Would be curious if you dont mind setting up an app with all of the options and hitting the endpoint. My understanding and reading of the code are that the two fields are not returned here (for some odd reason we decided that developer mode was an app setting and not a service setting so its define in the app/sync handler). That is a different endpoint and I think interestingly one that is not documented on here. 

TLDR, if you hit this endpoint: `"/groups/%s/apps/%s/sync/config"` you will get the following data: 

```
type syncData struct {
	DevelopmentModeEnabled bool   `json:"development_mode_enabled"`
	ServiceID              string `json:"service_id,omitempty"`
}
```

tkaye407: Confirmed this is correct. Reading the ticket I am curious if these fields were seen in the API? The weird thing (which I totally get) is that these fields are a part of the sync.json field in the cli download, so I am wondering if that is what was being confused cc @lindseymoore  / @nlarew

source/openapi-admin-v3.yaml
lindseymoore: Thank you!

source/openapi-admin-v3.yaml
lindseymoore: Thanks for checking and for pointing out the difference between the service config and the app's sync config. I can remove these two options from this API reference since they are not returned @tkaye407

source/openapi-admin-v3.yaml
```diff
@@ -7203,6 +7211,18 @@ components:
             names on **any collection**.
           items:
             type: string
+        indexed_queryable_fields_names:
+          type: array
+          description: |-
+            A list of field names to use as the indexed queryable field. While this property is an array, Sync 
+            currently supports only one one indexed queryable field. Therefore, this array may contain at most
```
cbullinger, line 7218:
```suggestion
            currently supports only one indexed queryable field. Therefore, this array may contain at most 
```

source/openapi-admin-v3.yaml
```diff
@@ -7203,6 +7211,18 @@ components:
             names on **any collection**.
           items:
             type: string
+        indexed_queryable_fields_names:
+          type: array
+          description: |-
+            A list of field names to use as the indexed queryable field. While this property is an array, Sync
```
cbullinger, line 7217: Maybe a link would be useful here?
```suggestion
            A list of field names to use as the [indexed queryable field](https://www.mongodb.com/docs/atlas/app-services/sync/configure/sync-settings/#queryable-fields). Although this property is an array, Sync 
```

source/sync/config-table.rst
```diff
@@ -89,6 +89,11 @@
      - A list of field names to use as :ref:`global queryable fields
        <queryable-fields>`.
 
+   * - | ``queryable_fields_version``
```
cbullinger, line 92: should the `config.json` example be updated to include this field? ...also, what even is the version of a queryable field?

cbullinger (approved):

source/sync/config-table.rst
lindseymoore: According to @jonatlee (thanks!), we should remove `queryable_fields_version` as it's not editable by the user. Removing it

jonatlee: Yeah for some more context, it is an internal field that we use to track when people make changes to their queryable fields. we shouldn't expose it to end users (we intentionally ignore any import requests coming from the cli with it set)

PR #714, opened 2024-03-27
DOCSP-38031 Updates to Authentication Provider Metadata

lindseymoore:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-38031

- [Define User Metadata](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-38031/users/custom-metadata/#custom-user-data): add notes about changes due to user metadata caching

### Reminder Checklist

  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

**Authenticate and Manage Users**
  - Define User Metadata: Add notes for updates to handling authentication user metadata.

### Review Guidelines

### DIFF
--- a/source/users/custom-metadata.txt
+++ b/source/users/custom-metadata.txt
@@ -4,7 +4,12 @@
 Define Custom User Metadata
 ===========================
 
-.. default-domain:: mongodb
+.. facet::
+   :name: genre 
+   :values: procedure
+
+.. meta:: 
+   :description: Learn how to define custom user metadata and read user metadata from authentication providers. 
 
 .. contents:: On this page
    :local:
@@ -74,8 +79,7 @@ When you use custom user data, keep the following things in mind:
   authentication access token. If a user's metadata is updated after the 
   access token is issued, requests using the previously created access 
   token will not have the newly updated metadata. Their user metadata 
-  in that session does not update until they refresh their access 
-  token or re-authenticate.
+  will update when they refresh their access token or re-authenticate.
 
 Create and Manage Custom User Data
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -732,6 +736,8 @@ Configure Authentication Provider Metadata
     providers. For example, both JWT and API Key authentication providers have 
     a ``name`` metadata field. There will be a delay to update the ``name`` field 
     in the user's table if the user authenticates with JWT and then with an API key. 
+    The request is always guaranteed to have the most up-to-date metadata associated 
+    with the authentication provider used.
 
 Access User Metadata from a Client Application
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/users/custom-metadata.txt
+++ b/source/users/custom-metadata.txt
@@ -728,6 +728,12 @@ Configure Authentication Provider Metadata
   in outdated metadata reflected in the user's table and for requests 
   using authentication providers.
 
+.. note::
+
+  For users using multiple authentcation providers and switching between 
+  authentcation providers to make requests, it may take up to 30 minutes for 
+  custom user data to update.
+
 Access User Metadata from a Client Application
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/source/users/custom-metadata.txt
+++ b/source/users/custom-metadata.txt
@@ -721,18 +721,17 @@ Configure Authentication Provider Metadata
          If you successfully configure the provider's metadata
          fields, App Services returns a ``204`` response.
 
-.. note:: Users Should Re-authenticate after Switching Between Authentication Providers
+.. note:: Tips for Users with Multiple Linked Authentication Providers  
 
-  To ensure the most up-to-date metadata, users should re-authenticate after 
-  switching to a different authentication provider. Failure to do so may result 
-  in outdated metadata reflected in the user's table and for requests 
-  using authentication providers.
-
-.. note::
-
-  For users with multiple authentcation providers, requests can be made using each of the 
-  authentication providers. The request will have the most up to date metadate associated with
-  the authentcation provider. However, there may be a 30 minute delay.  
+  - To ensure the most up-to-date metadata, users should re-authenticate after 
+    switching to a different authentication provider. Failure to do so may result 
+    in outdated metadata reflected in the user's table and for requests 
+    using authentication providers.
+  - If a user switches between authentication providers, there may be a 30 minute delay 
+    to update metadata that has the same field name for both authentication 
+    providers. For example, both JWT and API Key authentication providers have 
+    a ``name`` metadata field. There will be a delay to update the ``name`` field 
+    in the user's table if the user authenticates with JWT and then with an API key. 
 
 Access User Metadata from a Client Application
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


source/users/custom-metadata.txt
```diff
@@ -71,10 +76,10 @@ When you use custom user data, keep the following things in mind:
...
 - **Custom Data May Be Stale**: A user's custom data is *sourced* from a
   MongoDB collection but is stored in and read from a user's
-  authentication access token. If a user has a valid access token when
-  the underlying document changes, their custom data in that session
-  does not update until they refresh their access token or
-  re-authenticate.
+  authentication access token. If a user's metadata is updated after the 
+  access token is issued, requests using the previously created access 
+  token will not have the newly updated metadata. Their user metadata 
+  will update when they refresh their access token or re-authenticate.
```
tahegde, line 82: [Custom user data](https://www.mongodb.com/docs/atlas/app-services/users/custom-metadata/#custom-user-data) and [user metadata from authentication providers](https://www.mongodb.com/docs/atlas/app-services/users/custom-metadata/#authentication-provider-metadata) are two separate things so we would _not_ want to overwrite this. Instead we would want to just add this section you wrote most probably to [this section](https://www.mongodb.com/docs/atlas/app-services/users/custom-metadata/#authentication-provider-metadata)

source/users/custom-metadata.txt
```diff
@@ -432,6 +437,14 @@ the user's account. These metadata fields vary depending on provider.
        <custom-jwt-authentication-configuration-metadata-fields>`
        configuration.
 
+.. important:: Security and Authentication Provider Metadata 
+
+  Authentication provider metadata is externally-defined by clients and
```
tahegde, line 442: [nit] 
```suggestion
  Authentication provider metadata can be externally-defined by clients and
```

source/users/custom-metadata.txt
```diff
@@ -712,6 +725,19 @@ Configure Authentication Provider Metadata
          If you successfully configure the provider's metadata
          fields, App Services returns a ``204`` response.
 
+.. note:: Tips for Users with Multiple Linked Authentication Providers  
+
+  - To ensure the most up-to-date metadata, users should re-authenticate after 
+    switching to a different authentication provider. Failure to do so may result 
+    in outdated metadata reflected in the user's table and for requests
```
tahegde, line 732: This is a table with a list of all the users so I don't think this should be possessive!
```suggestion
    in outdated metadata reflected in the Users Table and for requests 
```

source/users/custom-metadata.txt
```diff
@@ -712,6 +725,19 @@ Configure Authentication Provider Metadata
...
+    switching to a different authentication provider. Failure to do so may result 
+    in outdated metadata reflected in the user's table and for requests 
+    using authentication providers.
+  - If a user switches between authentication providers, there may be a 30 minute delay 
+    to update metadata that has the same field name for both authentication 
+    providers. For example, both JWT and API Key authentication providers have 
+    a ``name`` metadata field. There will be a delay to update the ``name`` field 
+    in the user's table if the user authenticates with JWT and then with an API key. 
+    The request is always guaranteed to have the most up-to-date metadata associated 
+    with the authentication provider used.
```
tahegde, line 740: I don't think this example is necessary since the case is an edge case and there'd have to be other factors involved for this data to become stale--we don't want to confuse users unnecessarily. Also JWT authentication isn't guaranteed to have a "name" field. 

This also makes it seem like we'll only ever have stale metadata regarding the conflicting field name when in fact, other metadata could possibly be stale (especially if we end up expanding on our caching in the future).
```suggestion
  - If a user switches between authentication providers, there may be a 30 minute delay in propagating metadata to the Users Table and requests using authentication providers. The request is always guaranteed to have the most up-to-date metadata associated with the current authentication provider used.
```

tahegde (changes requested):
Thanks so much for doing this!! Let me know if you have any questions about any of my comments!

source/users/custom-metadata.txt
```diff
@@ -432,6 +437,14 @@ the user's account. These metadata fields vary depending on provider.
        <custom-jwt-authentication-configuration-metadata-fields>`
        configuration.
 
+.. important:: Security and Authentication Provider Metadata 
+
+  Authentication provider metadata is externally-defined by clients and
+  external authentication providers and should be regarded with caution. You should 
+  not rely solely on authentication provider metadata for security-related 
+  decisions, such as using this metadata in rule expressions for data access 
+  permssions.
```
tahegde, line 446: [typo]
```suggestion
  permissions.
```

source/users/custom-metadata.txt
lindseymoore: Do you know whether the table name is ``user`` or ``users``? Would like to rephrase user table to 'user collection', as we don't typically refer to data as being stored in tables in the docs

tahegde: ah when I say "Users Table", I mean an actual table component in our UI that contains a list of users for the admin. But actually, it might be clearer to say the "Users Page" instead if you're good with that!

lindseymoore: ok makes sense, thank you! Updated the wording, let me know what you think!

source/users/custom-metadata.txt
```diff
@@ -713,6 +733,18 @@ Configure Authentication Provider Metadata
...
 
+
+.. note:: Tips for Users with Multiple Linked Authentication Providers  
+
+  - To ensure the most up-to-date metadata, users should re-authenticate after 
+    switching to a different authentication provider. Failure to do so may result 
+    in outdated metadata reflected in the :guilabel:`Users` table in 
+    the UI's :guilabel:`App Users` page and for requests using authentication providers.
+  - If a user switches between authentication providers, there may be a 30 minute delay 
+    in propagating metadata to the :guilabel:`Users` table and for requests using
```
krollins-mdb, line 744: Minor suggestion for concision and avoid internal implementation detail that's not relevant to users.
```suggestion
  - If a user switches between authentication providers, it may take 30 minutes to
     to propagate metadata.
```

It's not clear to me how "and for requests using authentication providers" relates. Is there also a 30 minute delay in metadata there? If so, the next sentence seems to contradict that.

krollins-mdb (approved):
LGMT with one suggestion and a question.

source/users/custom-metadata.txt
lindseymoore: Removed the contradicting bit! Left the Users table part, as the engineer wanted to highlight that detail specifically to users.

tahegde: Yeah so the original message we were trying to convey was, "Each request is guaranteed to have the most up to date metadata associated with their auth provider, however changes can take up to 30 minutes to propagate to the users table and requests using other authentication providers." just to make it clear that authenticating with a _different_ authentication provider may not have the metadata specified in the request made prior with the original authentication provider used. 

With how this PR currently words this, I'd be good with going with @krollins-mdb suggestion and removing the "Users Table" part out of this as well!

source/users/custom-metadata.txt
```diff
@@ -589,129 +609,142 @@ Configure Authentication Provider Metadata
...
+
+
+.. note:: Tips for Users with Multiple Linked Authentication Providers  
 
+  - To ensure the most up-to-date metadata, users should re-authenticate after 
+    switching to a different authentication provider. Failure to do so may result 
+    in outdated metadata reflected in the :guilabel:`Users` table in 
+    the UI's :guilabel:`App Users` page and for requests using authentication providers.
+  - If a user switches between authentication providers, it may take up to 30 minutes 
+    to propagate metadata to the :guilabel:`Users` table. The request is always guaranteed to have the
```
tahegde, line 746: after the comment above, can we just actually remove the "Users table" bit? Since the above bullet has both "Users Table" and "requests using authentication methods", I don't want this next bullet to only mention one of those (when it does apply to both but we're not including it). 
```suggestion
  - If a user switches between authentication providers, it may take up to 30 minutes 
   for metadata to propagate. The request is always guaranteed to have the 
```

PR #728, opened 2024-04-12
(DOCSP-38851): Add a Manage Edge Servers page

dacharyc:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-38551

- [Manage Edge Servers](https://preview-mongodbdacharyc.gatsbyjs.io/atlas-app-services/DOCSP-38551/edge-server/manage-edge-servers/): New page describing how to manage Edge Servers in Atlas.

Two things to note:
- I've added placeholders to the "Configure" page for work that will come in PR #733 . There will be Snooty autobuilder failures here since those refs aren't present here.
- I've added placeholders in this page for Admin API details that will come in a later PR.

TODOS
These todos should carry over to the feature branch - they won't be resolved in this PR:
- [ ] Add info about the new Admin API endpoints to the Manage Edge Servers page once the new endpoints are documented
- [ ] Add links to the ref placeholders on the Manage Edge Servers page once the new pages are added
  - [ ] Metrics
  - [ ] Logs

### Reminder Checklist

  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

- **Edge Server**
  - Manage Edge Servers: New page describing how to manage Edge Servers in Atlas.

### Review Guidelines

### DIFF
--- a/source/edge-server/manage-edge-servers.txt
+++ b/source/edge-server/manage-edge-servers.txt
@@ -90,10 +90,7 @@ Services.
 
       .. include:: /includes/create-service-data-services.rst
 
-      .. important:: Enable Device Sync
-
-         Make sure to enable Device Sync, even if you choose an ``Edge Server``
-         Service.
+      .. include:: /includes/edge-server-enable-device-sync.rst
 
    .. tab:: Create an App Service
       :tabid: app-service
@@ -102,10 +99,7 @@ Services.
 
       When you choose your App type, select ``Edge Server`` or ``Device Sync``. 
 
-      .. important:: Enable Device Sync
-
-         Make sure to enable Device Sync, even if you choose an ``Edge Server``
-         Service.
+      .. include:: /includes/edge-server-enable-device-sync.rst
 
 .. _create-edge-server-instance:
 
--- a/source/edge-server/manage-edge-servers.txt
+++ b/source/edge-server/manage-edge-servers.txt
@@ -56,22 +56,20 @@ There are two types of services that enable you to use Edge Server:
   to MongoDB developers. The Edge Server Service rejects incoming Atlas Device
   SDK connections.
 - Device Sync Service, adding Edge Servers: use this option if you want to 
-  develop client apps using Atlas Device SDK. This option involves more strict 
-  schema enforcement, which the SDKs use to automatically sync objects 
-  across devices and handle conflict resolution.
+  develop client apps using Atlas Device SDK and MongoDB Drivers and tools. 
+  This option involves more strict schema enforcement, which the SDKs use to 
+  automatically sync objects across devices and handle conflict resolution.
 
 Regardless of which type of service you choose, Edge Servers use Device Sync 
-to sync data between the Edge Server instance and Atlas. Even when you choose 
-an Edge Server Service, you must first enable Device Sync before you can create 
-Edge Server instances.
+to sync data between the Edge Server instance and Atlas.
 
-.. important:: You can't add Atlas Device SDK Apps later to an Edge Server Service
+.. important:: You can't add Atlas Device SDK later to an Edge Server Service
 
    You can always add Edge Servers to a Device Sync Service. But you can't use
    Atlas Device SDK connections with an Edge Server Service. 
    
    Users are not portable between services. If you start with an Edge Server 
-   Service, but later want to add an Atlas Device SDK application, there is no 
+   Service, but later want to develop apps using Atlas Device SDK, there is no 
    way to transfer any users in your Edge Service to a new Device Sync Service.
 
 You can create an Edge Server-compatible service in Data Services or App 
@@ -90,29 +88,42 @@ Services.
 
       .. include:: /includes/create-service-data-services.rst
 
-      .. include:: /includes/edge-server-enable-device-sync.rst
-
    .. tab:: Create an App Service
       :tabid: app-service
 
       Refer to the documentation to :ref:`create-app`.
 
-      When you choose your App type, select ``Edge Server`` or ``Device Sync``. 
+      When you choose your App type, select ``Device Sync``. You can add
+      Edge Server to your Device Sync App after you configure Device Sync.
 
-      .. include:: /includes/edge-server-enable-device-sync.rst
+      [PLACEHOLDER: Add steps for adding Edge Server to a Device Sync App.
+      It's not 100% clear to me what the flow is from the Figma designs.
+      Populate the rest of this procedure when we are able to test this flow.]
 
 Supported Authentication Providers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Edge Server supports a subset of authentication providers:
+Edge Server supports a subset of authentication providers. Supported 
+authentication providers vary depending on whether you're connecting with
+Atlas Device SDK or using a MongoDB URI to connect with a MongoDB Driver 
+or tool.
+
+You must enable at least one of these providers to enable clients to connect
+to the Edge Server.
+
+Supported Atlas Device SDK Authentication Providers
+```````````````````````````````````````````````````
 
 - :ref:`anonymous-authentication`
 - :ref:`email-password-authentication`
 - :ref:`custom-jwt-authentication`
 - :ref:`api-key-authentication`
 
-You must enable at least one of these providers to enable clients to connect
-to the Edge Server.
+Supported MongoDB URI Authentication Providers
+``````````````````````````````````````````````
+
+- :ref:`anonymous-authentication`
+- :ref:`email-password-authentication`
 
 .. _create-edge-server-instance:
 
@@ -132,15 +143,8 @@ registration key.
 
          .. step:: Add a server
 
-            Click the :guilabel:`Add a server` button to walk through the 
-            process of configuring your new Edge Server instance.
-
-            If this is the first server in your service, you're prompted to 
-            select a cluster, connect a data source, and optionally specify 
-            which collections to sync to the Edge Server. This configuration
-            applies to all Edge Servers in this service.
-
-            For subsequent servers, you're prompted to name you Edge Server.
+            Click the :guilabel:`Add a server`. You're prompted to name your 
+            Edge Server.
 
             The UI provides an installation script, which you can copy and 
             paste into the terminal on the device where you want to install 
@@ -379,33 +383,42 @@ Edge Server Service
 ~~~~~~~~~~~~~~~~~~~
 
 When you're using an Edge Server Service, the service automatically updates
-properties in a collection when you make changes to the documents you sync.
-However, some types of changes cause an Edge Server instance to reboot and
-wipe itself.
+fields in the schema when you make changes to the documents you sync.
+
+If you update your schema manually using the UI, and those schema changes
+trigger a schema version change, this causes the Edge Server instance to
+reboot and wipe itself.
 
 If you're adding a new collection you haven't synced before, you must define
-a new schema that contains only the collection's name.
+a new schema.
 
 Add a Collection
 ````````````````
 
 When you want to add a collection to an Edge Server Service, you must add
-a schema containing only the collection's name. Do not add any properties
-to the schema. The Edge Server Service automatically infers the properties
-from the shape of the documents you sync or the documents in the collection.
+a schema containing:
+
+- The collection's name
+- An ``_id`` field that is an ``ObjectID`` type.
+
+The Edge Server Service automatically infers the other fields from the shape 
+of the documents you sync or the documents in the collection.
 
 For information about how to add a schema, refer to :ref:`enforce-a-schema`.
 
-Change Properties in a Collection
-`````````````````````````````````
+Change Fields in a Document
+```````````````````````````
 
-When you change the properties in a collection using an Edge Server Service,
-the service automatically changes the schema to match the new properties.
+When you change the fields in a document using an Edge Server Service,
+the service automatically changes the schema to add new fields.
 
-Some changes are considered "breaking" schema changes. When a "breaking"
-schema change occurs, any Edge Server instance with the old schema reboots
-and wipes itself. The wipe process erases any unsynced data. Edge Server will 
-get the ability to recover local unsynced data in a future release.
+Some changes that you can make through the UI are considered "breaking" 
+schema changes. When a "breaking" schema change occurs, any Edge Server 
+instance with the old schema reboots and wipes itself. The wipe process 
+erases any unsynced data.
+
+Edge Server will get the ability to recover local unsynced data in a future 
+release.
 
 For details about which types of changes are considered "breaking" changes,
 refer to :ref:`breaking-change-quick-reference`.
@@ -418,6 +431,14 @@ updating a data model.
 
 When a "breaking" schema change occurs, any Edge Server instance with the old 
 schema reboots and wipes itself. The wipe process erases any unsynced data. 
+
+If that data was written by an Atlas Device SDK client, the client itself
+can handle data recovery. This enables local unsynced data written by the 
+client to eventually make its way to Atlas.
+
+If the data was written by a MongoDB Driver or tool using the wireprotocol,
+the unsynced data is unrecoverable.
+
 Edge Server will get the ability to recover local unsynced data in a future 
 release.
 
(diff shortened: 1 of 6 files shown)


source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+an Edge Server Service, you must first enable Device Sync before you can create 
+Edge Server instances.
+
+.. important:: You can't add Atlas Device SDK Apps later to an Edge Server Service
+
+   You can always add Edge Servers to a Device Sync Service. But you can't use
+   Atlas Device SDK connections with an Edge Server Service. 
+   
+   Users are not portable between services. If you start with an Edge Server 
+   Service, but later want to add an Atlas Device SDK application, there is no
```
rukumar333, line 74: Same thing here, do we want to call this an application? Is this referring to a mobile application?

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+.. important:: You can't add Atlas Device SDK Apps later to an Edge Server Service
+
+   You can always add Edge Servers to a Device Sync Service. But you can't use
+   Atlas Device SDK connections with an Edge Server Service. 
+   
+   Users are not portable between services. If you start with an Edge Server 
+   Service, but later want to add an Atlas Device SDK application, there is no 
+   way to transfer any users in your Edge Service to a new Device Sync Service.
+
+You can create an Edge Server-compatible service in Data Services or App
```
rukumar333, line 77: As a note, I think the `Edge Server` service can only be created via Data Services correctly. Do you think that is covered with the information below?

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+   
+   Users are not portable between services. If you start with an Edge Server 
+   Service, but later want to add an Atlas Device SDK application, there is no 
+   way to transfer any users in your Edge Service to a new Device Sync Service.
+
+You can create an Edge Server-compatible service in Data Services or App 
+Services. 
+
+- You might select Data Services when you just want to get started building
+  with Edge Server or Device Sync and Edge Server, and don't need other
```
rukumar333, line 81: Do we want this to say "with an Edge Server" or "with Edge Servers"? I guess it makes sense to maybe keep this as is since Edge Server is the product name but was just a question.

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+      .. include:: /includes/create-service-data-services.rst
+
+      .. include:: /includes/edge-server-enable-device-sync.rst
+
+   .. tab:: Create an App Service
+      :tabid: app-service
+
+      Refer to the documentation to :ref:`create-app`.
+
+      When you choose your App type, select ``Edge Server`` or ``Device Sync``.
```
rukumar333, line 100: Yeah we might want to change this since I think we can only create an Edge Server service via the Data Services page (@sudssm want to confirm this).

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+
+Edge Server supports a subset of authentication providers:
+
+- :ref:`anonymous-authentication`
+- :ref:`email-password-authentication`
+- :ref:`custom-jwt-authentication`
+- :ref:`api-key-authentication`
+
+You must enable at least one of these providers to enable clients to connect
+to the Edge Server.
```
rukumar333, line 115: Is it worth calling out here that Device Sync clients can connect via any of these authentication methods but MongoDB driver access is only via anonymous auth or email-password (this is only for public preview).

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+
+.. _create-edge-server-instance:
+
+Create an Edge Server Instance
+------------------------------
+
+After creating an Edge Server Service, you can add Edge Servers to it. The
+service can handle multiple Edge Servers, each of which has its own 
+registration key.
```
rukumar333, line 124: Do we want to say its own API key here? That is what the Edge Server actually uses to authenticate with the Cloud, the registration key is only used to retrieve the API key. But also, the customer will only see the registration key and not the API key so curious what you think would be better?

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+         .. step:: Add a server
+
+            Click the :guilabel:`Add a server` button to walk through the 
+            process of configuring your new Edge Server instance.
+
+            If this is the first server in your service, you're prompted to 
+            select a cluster, connect a data source, and optionally specify 
+            which collections to sync to the Edge Server. This configuration
+            applies to all Edge Servers in this service.
+
```
rukumar333, line 142: The cluster used is only when setting up the Edge Server service. When going to add an Edge Server (whether the first or not) while already in the service/application page doesn't require setting this up. It will use the same cluster that device sync is using.

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+            instance to your service. You must provide the the App's internal 
+            ``appId`` hex string:
+
+            .. code-block:: shell
+
+               INSERT API DETAILS HERE
+
+            This returns <INSERT RETURN DETAILS HERE>.
+
+         .. step:: Install the Edge Server code
```
rukumar333, line 181: Oh I don't think the admin endpoint returns the install instructions.

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+----------------------------------
+
+The process to update a data model varies depending on whether you're using
+an Edge Server Service or a Device Sync Service with Edge Server.
+
+Edge Server Service
+~~~~~~~~~~~~~~~~~~~
+
+When you're using an Edge Server Service, the service automatically updates
+properties in a collection when you make changes to the documents you sync.
```
rukumar333, line 382: I am not sure if we want to use app services terminology here, do we want to say that the fields in the schema are automatically updated?

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+However, some types of changes cause an Edge Server instance to reboot and
```
rukumar333, line 383: The only schema changes that would require a reboot and wipe would be schema changes which result in a schema version change which are only changes that could be triggered by the customer updating their schema via the UI. Maybe we should make this clear so that the customer doesn't think that the Edge Server can reboot + wipe on its own due to a schema change unless it is the customer modifying the schema?

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+wipe itself.
+
+If you're adding a new collection you haven't synced before, you must define
+a new schema that contains only the collection's name.
```
rukumar333, line 387: 😬 They will also need to define the _id field as an ObjectID. On that note, I think we need to clarify somewhere that the Edge Server service currently only supports documents with _id fields that are ObjectIDs, other types are not currently supported (again this is only for the Edge Server service).

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+
+Add a Collection
+````````````````
+
+When you want to add a collection to an Edge Server Service, you must add
+a schema containing only the collection's name. Do not add any properties
```
rukumar333, line 393: Same thing here, the _id field needs to be set to an object id.

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+refer to :ref:`breaking-change-quick-reference`.
+
+Device Sync Service
+~~~~~~~~~~~~~~~~~~~
+
+An Edge Server running in a Device Sync Service uses Device Sync rules for
+updating a data model.
+
+When a "breaking" schema change occurs, any Edge Server instance with the old 
+schema reboots and wipes itself. The wipe process erases any unsynced data.
```
rukumar333, line 420: FWIW, unsynced data written by a Device Sync client will eventually make its way to the Cloud server even on a wipe since the Device Sync client itself can handle data recovery.

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,485 @@
...
+
+A few conditions can cause an Edge Server instance to enter a ``Failed`` state:
+
+- :ref:`edge-server-exceeds-client-maximum-offline-time`
+- :ref:`edge-server-more-than-one-major-version-behind`
+- :ref:`edge-server-token-invalid`
+
+.. _edge-server-exceeds-client-maximum-offline-time:
+
+Edge Server is Offline Longer than Client Maximum Offline Time
```
rukumar333, line 439: It is worth noting, we are currently enforcing that the Edge Server cannot be offline for longer than 60 days no matter what this setting is. So if the setting is nil or is 120 days, we will enforce 60 days.

source/edge-server/manage-edge-servers.txt
dacharyc: Yeah, since registration key is the language used in the UI, that's the language I've chosen here for the docs. On the "Configure" page, I cover that the registration key is exchanged for an API Key.

I don't know how much into the weeds we want to get around these details, so I'll defer to you and Sudarshan about that. If it seems important to use "API Key" for security purposes or otherwise, I'm happy to change it.

source/edge-server/manage-edge-servers.txt
dacharyc: The other page that I'm linking out to here has the install instructions. I've added those in PR #733 .

source/edge-server/manage-edge-servers.txt
dacharyc: I've added some details below that cover this - thanks for calling it out!

source/edge-server/manage-edge-servers.txt
dacharyc: I believe so. This is why I said "Edge Server-compatible service" instead of "Edge Server Service". I was hoping to cover both Device Sync with Edge Server and just Edge Server. If it seems confusing, though, I'm happy to change it.

source/edge-server/manage-edge-servers.txt
dacharyc: Yeah, I went with "Edge Server" as the product name here, but am willing to change it if it seems confusing.

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,518 @@
...
+.. _edge-server-pending:
+
+Pending Edge Server Instances
+`````````````````````````````
+
+When an Edge Server instance has not yet connected to Atlas for the first time, 
+the instance's :guilabel:`Health` status is :guilabel:`Pending`.
+
+While an Edge Server instance is in :guilabel:`Pending` state, the detail
+view includes:
```
cbullinger, line 266: are long bullet lists the the best way to communicate this? it's difficult to parse the two sub-sections and their content (are we revisiting this section once the UI is finalized?)

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,518 @@
...
+
+After you have connected an Edge Server instance to Atlas for the first time, 
+the instance's :guilabel:`Health` status changes to :guilabel:`Active`. 
+
+If the Edge Server instance encounters a fatal error, its :guilabel:`Health`
+status changes to :guilabel:`Failed`.
+
+The details available for active or failed instances include:
+
+- A button you can use to rename the instance.
```
cbullinger, line 292: same

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,518 @@
...
+storage required in the backing Atlas cluster. The longer the client maximum
+offline time, the more storage Atlas requires to keep historical records that
+support Sync.
+
+If an Edge Server instance is offline longer than this setting, this is an 
+unrecoverable error. You must wipe and restart the instance. Alternately, when 
+you bring the instance back online, it will be wiped and restarted 
+automatically.
+
+**This process erases all local unsynced data on the Edge Server instance.**
```
cbullinger, line 481: the bold formatting here makes it look like it's a new sub-heading, rather than an emphasized line. might be better suited as an admonition?

cbullinger (approved):
Some suggestions. Still not sure how best to talk about this stuff, though ☹️

source/edge-server/manage-edge-servers.txt
rukumar333: Ah nope, makes sense!

source/edge-server/manage-edge-servers.txt
```diff
@@ -90,29 +88,42 @@ Services.
...
 - :ref:`email-password-authentication`
 - :ref:`custom-jwt-authentication`
 - :ref:`api-key-authentication`
 
-You must enable at least one of these providers to enable clients to connect
-to the Edge Server.
+Supported MongoDB URI Authentication Providers
+``````````````````````````````````````````````
+
+- :ref:`anonymous-authentication`
```
rukumar333, line 127: @sudssm Hoping to get your input here, do we want to represent the --insecure-disable-auth flow like this here? Or do you think we should drop this part?

source/edge-server/manage-edge-servers.txt
rukumar333: Yeah I think this is good as is then, thanks!

source/edge-server/manage-edge-servers.txt
```diff
@@ -460,6 +481,13 @@ will have the option to recover local data when this error occurs.
 You can change the length of the client maximum offline time in the 
 :ref:`sync-settings`.
 
+.. important:: Edge Server Offline Limit is 60 Days
+
+   An Edge Server instance cannot be offline longer than 60 days, no matter
+   what setting you set for the client maximum offline time. If the setting
+   is ``nil`` or greater than 60 days, the Edge Server offline limit is
```
rukumar333, line 493: Oh we should say "If the setting is unset" I think instead of nil here. My bad, I was using Golang terminology here.

source/edge-server/manage-edge-servers.txt
rukumar333: Maybe good to get Ian's input here, I don't know the right answer. I can raise this to him in the channel.

source/edge-server/manage-edge-servers.txt
dacharyc: I've updated this to reframe as "actions you can do". If you've got other ideas about how to present this information, I'm happy to revise.

source/edge-server/manage-edge-servers.txt
```diff
@@ -0,0 +1,513 @@
...
+
+      [PLACEHOLDER: Add steps for adding Edge Server to a Device Sync App.
+      It's not 100% clear to me what the flow is from the Figma designs.
+      Populate the rest of this procedure when we are able to test this flow.]
+
+.. important:: Required Field
+
+   Documents that you want to sync with Edge Server must contain an ``_id`` 
+   field of type ``ObjectID``. Any documents without this field cannot sync 
+   to an Edge Server.
```
rukumar333, line 109: This is only true if you are using the Edge Server service. For device sync, this is not required.

rukumar333 (approved):
LGTM!

source/edge-server/manage-edge-servers.txt
dacharyc: We state [in the docs](https://www.mongodb.com/docs/atlas/app-services/sync/data-model/create-a-schema/#std-label-create-sync-schema) that Device Sync requires a primary key `_id` property, but I guess it doesn't have to be an `ObjectID` - our docs state it could also be a `string` or an `int`. So is that the difference - for Edge Server it has to be an `ObjectID` but for Device Sync, there must be an `_id` of a valid type? Or are the docs incorrect and `_id` isn't required for Sync?

rukumar333: Device Sync Service:
 _id field can be of type ObjectId, String, Long, Int, UUID

Edge Server Service:
_id field has to be of type ObjectId

rukumar333: I DM'ed as well and although the _id field actually isn't required for the edge server or device sync, we don't really encourage customers to do make it not-required so we can keep it in the docs that the _id field is required.

source/edge-server/manage-edge-servers.txt
dacharyc: I'm going to go ahead and merge this PR now to keep things moving along, but I'm happy to change this in the feature branch before publication if there's a different way you'd prefer to communicate these details.

PR #735, opened 2024-04-16
(DOCSP-38553): Update the 'Connect' page for Public Preview

dacharyc:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-38553

- [Connect to an Edge Server](https://preview-mongodbdacharyc.gatsbyjs.io/atlas-app-services/DOCSP-38553/edge-server/connect/): Update the page for the Public Preview. Add details about:
  - Roaming with Atlas Device SDK
  - Connecting via the wireprotocol with authentication
  - Connecting via Compass

Note: There is currently a placeholder in the "Connect with Compass" section, as I do not know what URI to use to connect with Compass. I'll update this section when I get the info from engineering, or when I can test it myself.

### Reminder Checklist

  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

- **Edge Server**
  - Connect to an Edge Server: Update the page for the Public Preview. Add details about:
    - Roaming with Atlas Device SDK.
    - Connecting via the wireprotocol with authentication.
    - Connecting via Compass.

### Review Guidelines

### DIFF
--- a/source/edge-server/connect.txt
+++ b/source/edge-server/connect.txt
@@ -72,7 +72,6 @@ connecting with an SDK:
 - :ref:`anonymous-authentication`
 - :ref:`email-password-authentication`
 - :ref:`api-key-authentication`
-- :ref:`custom-function-authentication`
 - :ref:`custom-jwt-authentication`
 
 .. _edge-server-device-sdk-roaming:
@@ -86,6 +85,10 @@ instance and Atlas.
 Roaming requires changing the base URL in the client's App configuration.
 For details, refer to the SDK documentation linked above.
 
+When an SDK user roams, the client app must perform a client reset upon
+connecting to the new server. For more information about client resets, refer 
+to :ref:`client-resets`.
+
 Roaming Users and Anonymous Authentication
 ``````````````````````````````````````````
 
@@ -96,10 +99,6 @@ local data across servers.
 
 For all other supported authentication providers, users persist and have 
 access to the expected data across servers.
-
-When an SDK user roams, the client app must perform a client reset upon
-connecting to the new server. For more information about client resets, refer 
-to :ref:`client-resets`.
 
 Example: SwiftUI Template App
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -239,8 +238,8 @@ You can connect to the Edge Server using standard MongoDB Drivers and tools.
 Clients use a specialized MongoDB URI connection string to connect to 
 Edge Server and send requests.
 
-The Wire Protocol server accepts these incoming connections, and is on by 
-default. You can connect to the host machine with a MongoDB URI.
+The Wire Protocol server accepts these incoming connections. You can connect 
+to the host machine with a MongoDB URI.
 
 This URI may vary depending on the authentication type. Broadly, it resembles
 this form:
@@ -295,6 +294,12 @@ Anonymous Connections
 
 You can enable anonymous connections and connect without specifying a username 
 or password. This option is disabled by default.
+
+.. important:: Disable Anonymous Connections in Production
+   
+   Enabling anonymous connections can streamline development by bypassing
+   authentication flows. This is only intended as a development aid; do not 
+   bypass authentication flows in production environments.
 
 .. procedure::
 
--- a/source/edge-server/connect.txt
+++ b/source/edge-server/connect.txt
@@ -335,15 +335,6 @@ or password. This option is disabled by default.
 
             mongodb://localhost:27021
 
-Supported Authentication Providers
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Edge Server currently supports a subset of authentication providers when
-connecting over the Wire Protocol:
-
-- :ref:`anonymous-authentication`
-- :ref:`email-password-authentication`
-
 Supported MongoDB APIs
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/source/edge-server/connect.txt
+++ b/source/edge-server/connect.txt
@@ -63,6 +63,26 @@ documentation for your preferred SDK:
 - React Native SDK: :ref:`react-native-connect-to-mongodb-realm-backend-app`
 - Swift SDK: :ref:`ios-init-appclient`
 
+.. _edge-server-device-sdk-roaming:
+
+Roaming Between Edge Servers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+SDK clients can roam between Edge Servers, or between an Edge Server and Atlas.
+
+Roaming requires changing the base URL in the client's App configuration.
+For details, refer to the SDK documentation linked above.
+
+Roaming Users and Anonymous Authentication
+``````````````````````````````````````````
+
+When roaming between Edge Servers or an Edge Server and Atlas, anonymous users
+do not persist across servers. Atlas creates a new anonymous user for each 
+server the client connects to, and does not persist local data across servers.
+
+For all other supported authentication providers, users persist and have 
+access to the expected data across servers.
+
 Example: SwiftUI Template App
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -206,13 +226,43 @@ host machine using this MongoDB URI:
 
 .. code:: console
 
-   mongodb://EDGE_SERVER_HOSTNAME:27021
+   mongodb://<username>:<password>@<hostname>:<port>?authMechanism=PLAIN&appName=::<auth-type>
 
-The ``EDGE_SERVER_HOSTNAME`` is the public DNS of your Edge Server host. 
-For more information, refer to :ref:`edge-server-install-and-configure`.
+.. list-table::
+   :header-rows: 1
 
-You must make port 27021 accessible to the client through port mapping or
-firewall rules.
+   * - Element
+     - Description
+
+   * - ``<username>``
+     - The username of the Atlas user. For an 
+       :ref:`email/password user <email-password-authentication>`, this 
+       might look similar to ``first.last@example.com``.
+
+   * - ``<password>``
+     - The password of the Atlas user. This might look similar to 
+       ``myPassword123456``.
+
+   * - ``<hostname>``
+     - The public DNS of your Edge Server host. While in local development, 
+       this might be ``localhost``. Otherwise, it might look similar to
+       ``ec2-13-58-70-88.us-east-2.compute.amazonaws.com``. For more 
+       information, refer to :ref:`edge-server-install-and-configure`.
+
+   * - ``<port>``
+     - The Edge Server wireprotocol listen port. The default port is ``27021``.
+       For more details, refer to :ref:`edge-server-install-and-configure`.
+
+   * - ``authMechanism=PLAIN``
+     - The auth mechanism. Currently, the only supported value is ``PLAIN``.
+
+   * - ``appName``
+     - Optional string ``appName``. You may omit the value and/or key.
+
+   * - ``<auth-type>``
+     - The type of authentication provider. For example, ``local-userpass`` 
+       for email/password authentication. Currently, the only supported type
+       is ``local-userpass``.
 
 .. example::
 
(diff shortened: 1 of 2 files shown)


source/edge-server/connect.txt
```diff
@@ -201,27 +227,82 @@ You can connect to the Edge Server using standard MongoDB Drivers and tools.
 Clients use a specialized MongoDB URI connection string to connect to 
 Edge Server and send requests.
 
-The wire protocol server is on by default. You can connect to it via the 
-host machine using this MongoDB URI:
+The Wire Protocol server accepts these incoming connections, and is on by 
+default. You can connect to it via the host machine using this MongoDB URI:
```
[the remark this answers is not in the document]
dacharyc: Changing it, because I've also been called out for using it, but I do feel "via" is not a foreign word. It's in English language dictionaries and has been accepted into the lexicon. 😛

source/edge-server/connect.txt
```diff
@@ -201,27 +227,82 @@ You can connect to the Edge Server using standard MongoDB Drivers and tools.
...
 
 .. code:: console
 
-   mongodb://EDGE_SERVER_HOSTNAME:27021
+   mongodb://<username>:<password>@<hostname>:<port>?authMechanism=PLAIN&appName=::<auth-type>
```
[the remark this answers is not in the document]
dacharyc: I was torn about how to document this, since `appName` is (currently) completely optional and is actually ignored in the code path.

So this _could_ be:

`appName=<your-app-name>` or it could be omitted altogether. I punted and just went with the engineering example, but I'm open to your thoughts about how to make this clear in the example, or whether we should just omit it.

source/edge-server/connect.txt
```diff
@@ -201,27 +227,82 @@ You can connect to the Edge Server using standard MongoDB Drivers and tools.
...
+     - Description
+
+   * - ``<username>``
+     - The username of the Atlas user. For an 
+       :ref:`email/password user <email-password-authentication>`, this 
+       might look similar to ``first.last@example.com``.
+
+   * - ``<password>``
+     - The password of the Atlas user. This might look similar to 
+       ``myPassword123456``.
```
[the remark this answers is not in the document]
dacharyc: I added it to continue the pattern of providing examples, and because anecdotally some docs users report getting confused about whether to leave in things like `<>` when they see example strings. So when I use special characters to provide context, I try to always provide an example showing the proper formatting of the value that omits the special characters.

source/edge-server/connect.txt
```diff
@@ -63,12 +63,45 @@ documentation for your preferred SDK:
...
+
+Roaming Users and Anonymous Authentication
+``````````````````````````````````````````
+
+When roaming between Edge Server instances or an Edge Server instance and 
+Atlas, anonymous users do not persist across servers. Atlas creates a new 
+anonymous user for each server the client connects to, and does not persist 
+local data across servers.
+
+For all other supported authentication providers, users persist and have
```
sudssm, line 96: maybe worth noting that the device will be client reset when it roams?

source/edge-server/connect.txt
```diff
@@ -83,10 +116,10 @@ modifications you can make to the :github:`Swift template app
            - Value
 
          * - ``baseUrl``
-           - The public-accessible DNS of your Edge Server's host.
+           - The public-accessible DNS of your Edge Server instance host.
              This is ``http://``, or if TLS is enabled, ``https://``, 
-             followed by the same value you set in your Edge Server 
-             :file:`edge_config.json` ``baseURL`` field.
+             followed by the same value you set in your Edge Server
+             instance configuration.
```
sudssm, line 122: this is not somehting they'd set in their configuration, rather its the ip address exposed by their docker or kubernetes cluster (so maybe can describe this as their "container configuration" instead?

source/edge-server/connect.txt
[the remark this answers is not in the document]
sudssm: this is correct. appName is optional and can be omitted if you're using username password. if you use other auth providers (not yet supported) you'll need that extra field.

worth also noting that you can start up the edge server with `--insecure-disable-auth`, which will allow you to connect without specifying a user/pass (as before), at which point you can do `mongodb://EDGE_SERVER_HOSTNAME:27021`.

source/edge-server/connect.txt
```diff
@@ -170,58 +203,122 @@ modifications you can make to the :github:`Swift template app
...
-   mongodb://EDGE_SERVER_HOSTNAME:27021
+   mongodb://<username>:<password>@<hostname>:<port>?authMechanism=PLAIN&appName=::<auth-type>
+
+.. list-table::
+   :header-rows: 1
+
+   * - Element
+     - Description
+
+   * - ``<username>``
```
sudssm, line 250: note that email password users dont actually have to be email usernames, and indeed, this would complicate things because of the "@" having to be escaped. so I'd prefer guiding users toward non-email usernames

sudssm: or, we should give an example with the @ properly escaped

source/edge-server/connect.txt
```diff
@@ -170,58 +203,122 @@ modifications you can make to the :github:`Swift template app
...
+     - The Edge Server Wire Protocol listen port. The default port is ``27021``.
+       For more details, refer to :ref:`edge-server-install-and-configure`.
+
+   * - ``authMechanism=PLAIN``
+     - The auth mechanism. Currently, the only supported value is ``PLAIN``.
+
+   * - ``appName``
+     - Optional string ``appName``. You may omit the value and/or key.
+
+   * - ``<auth-type>``
```
sudssm, line 275: i think this is confusing in that this is not needed for local-userpass or anonymous. maybe we can just drop it for now and add it back when we support other providers?

source/edge-server/connect.txt
```diff
@@ -170,58 +203,122 @@ modifications you can make to the :github:`Swift template app
...
+.. _edge-server-compass-connect:
+
+Connect to the Edge Server with Compass
+---------------------------------------
+
+You can connect to the Edge Server instance using 
+:ref:`MongoDB Compass <compass-index>`. Compass enables you to explore the 
+documents syncing to Edge Server instance, and make updates to those documents.
+
+[PLACEHOLDER - not sure exactly what URI to use to connect with Compass -
```
sudssm, line 313: same URI as above for drivers. `mongodb://host:port` if anonymous, or `mongodb://user:pass@host:port` if userpass

sudssm (approved):

source/edge-server/connect.txt
dacharyc: Ok, I'm added an example using percent encoding to escape the `@`. If there's another way we'd prefer to show this, I'm happy to update.

source/edge-server/connect.txt
dacharyc: Since neither `appName` nor `<auth-type>` is required currently, I've just removed them both to simplify this. Happy to add them back when we require them.

source/edge-server/connect.txt
```diff
@@ -63,12 +63,49 @@ documentation for your preferred SDK:
...
+Supported Authentication Providers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Edge Server currently supports a subset of authentication providers when
+connecting with an SDK:
+
+- :ref:`anonymous-authentication`
+- :ref:`email-password-authentication`
+- :ref:`api-key-authentication`
+- :ref:`custom-function-authentication`
```
rukumar333, line 75: We do not support custom function authentication but we do support custom JWT support.

source/edge-server/connect.txt
```diff
@@ -63,12 +63,49 @@ documentation for your preferred SDK:
...
+
+.. _edge-server-device-sdk-roaming:
+
+Roaming Between Edge Servers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+SDK clients can roam between Edge Server instances, or between an Edge Server
+instance and Atlas.
```
rukumar333, line 83: Do we usually say Device SDK clients connect to Atlas? Or do we say they connect to app services?

source/edge-server/connect.txt
```diff
@@ -170,58 +207,155 @@ modifications you can make to the :github:`Swift template app
...
+---------------------------------------------------------
 
 You can connect to the Edge Server using standard MongoDB Drivers and tools.
 Clients use a specialized MongoDB URI connection string to connect to 
 Edge Server and send requests.
 
-The wire protocol server is on by default. You can connect to it via the 
-host machine using this MongoDB URI:
+The Wire Protocol server accepts these incoming connections, and is on by 
+default. You can connect to the host machine with a MongoDB URI.
```
rukumar333, line 242: I don't think we actually support disabling the Wire Protocol server currently, @sudssm is this something we want? I think we discussed this before.

source/edge-server/connect.txt
```diff
@@ -170,58 +207,155 @@ modifications you can make to the :github:`Swift template app
...
+~~~~~~~~~~~~~~~~~~~~~
+
+You can enable anonymous connections and connect without specifying a username 
+or password. This option is disabled by default.
+
+.. procedure::
+
+   .. step:: Configure Edge Server with the Relevant Flag
+
+      To enable anonymous connections, start the Edge Server with the
```
rukumar333, line 302: Should we add a note here that this is not recommended for production usage?

source/edge-server/connect.txt
rukumar333: Actually, I need to confirm this, custom function authentication might actually work. But to confirm, custom JWT does work.

source/edge-server/connect.txt
```diff
@@ -170,58 +207,155 @@ modifications you can make to the :github:`Swift template app
    .. step:: Run the App
```
cbullinger, line 207: I know it's not the critical portion of this page, but if you're so inclined, the `Adjust the App Transport Security Settings (Optional)` step would benefit from having sub-steps. e.g.

> #### Adjust the App Transport Security Settings (Optional)
>   
> If your app uses TLS, you can disregard this step. If you are developing an Edge Server app with TLS not enabled during development, you must complete this step.
>  
> For iOS, Apple disables insecure connections by default. You must add an exception for your public-accessible DNS to allow the client app to connect to it:
> 
> 1. Select your App target, and go to the :guilabel:`Signing & Capabilities` tab. 
> 2. Press the :guilabel:`+ Capability` button, and search for ``App Transport Security Settings``. Add this to your app.
> 3. Add an exception domain for your public-accessible DNS. This should be just the string domain, similar to: ``ec2-13-58-70-88.us-east-2.compute.amazonaws.com``.
> ...

source/edge-server/connect.txt
```diff
@@ -170,58 +207,155 @@ modifications you can make to the :github:`Swift template app
...
+
+Connect with MongoDB Compass
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can connect to the Edge Server instance using the MongoDB URI through 
+:ref:`MongoDB Compass <compass-index>`. Compass enables you to explore the 
+documents syncing to Edge Server instance, and make updates to those documents.
+
+Limitations
+```````````
```
cbullinger, line 354:
```suggestion
~~~~~~~~~~~
```
(assuming these are the headings we're sticking with)

source/edge-server/connect.txt
```diff
@@ -170,58 +207,155 @@ modifications you can make to the :github:`Swift template app
...
+Supported MongoDB APIs
+~~~~~~~~~~~~~~~~~~~~~~
 
 Edge Server currently supports a subset of the MongoDB APIs. For information 
 about the supported APIs, refer to :ref:`edge-server-mongodb`.
+
+.. _edge-server-compass-connect:
+
+Connect with MongoDB Compass
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
cbullinger, line 347:
```suggestion
Connect to the Edge Server with MongoDB Compass
-----------------------------------------------
```

source/edge-server/connect.txt
```diff
@@ -170,58 +207,155 @@ modifications you can make to the :github:`Swift template app
...
+
+.. _anonymous-connection-URI:
+
+Anonymous Connections
+~~~~~~~~~~~~~~~~~~~~~
+
+You can enable anonymous connections and connect without specifying a username 
+or password. This option is disabled by default.
+
+.. procedure::
```
cbullinger, line 298:
```suggestion
You might want to connect to the Edge Server without credentials while in development. You can enable anonymous connections and connect without specifying a username 
or password. This option is disabled by default. 

.. important:: Disable Anonymous Connections in Production

   Anonymous connections to the Edge Server instance are not recommended in production environments. 

.. procedure::
```
...or however we want to word this

source/edge-server/connect.txt
dacharyc: I don't think I've seen anything indicating that custom function auth is supposed to be supported - I think I just grabbed the wrong reference here. Updating it to custom JWT!

source/edge-server/connect.txt
dacharyc: I think this was language we added a while ago, as it seemed worth noting since we did not require authentication to connect via the Wire Protocol. But since that is no longer the case, I'll just remove the "on by default" language from here. If you decide later there's something we want to call out around this, I'm happy to update it.

source/edge-server/connect.txt
dacharyc: I initially had this as its own section. But it uses the same MongoDB URI to connect, so it didn't seem worth having another section for it and then referring people to this section for the connection string details. I left it in the intro because I wanted to call it out as a way to explore documents on the Edge Server, which seemed useful for folks who are in development and want to understand what's syncing to the Edge Server.

source/edge-server/connect.txt
dacharyc: We're inconsistent, but I have generally been hand-waving this and saying "Atlas" since we added the ability to create a Device Sync service in the Data Services tab.

rukumar333 (approved):

PR #751, opened 2024-04-30
DOCSP-38871 Add docs release notes for w/e April 26

lindseymoore:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-38871

### DIFF
--- a/docs-release-notes.md
+++ b/docs-release-notes.md
@@ -1,3 +1,21 @@
+# April 26, 2024
+
+## Tutorials
+
+- Change "Get Started" TOC entry to "Tutorials".
+- Flip the TOC construction for tutorial pages from "Device Sync Tutorial - " to " - Device Sync".
+- Individual tutorial pages:
+  - Update page titles to clearly indicate they're tutorials and include product names.
+  - Add meta descriptions and keywords to those that didn't have them.
+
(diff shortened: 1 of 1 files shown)


docs-release-notes.md
```diff
@@ -1,3 +1,21 @@
...
+  - Update page titles to clearly indicate they're tutorials and include product names.
+  - Add meta descriptions and keywords to those that didn't have them.
+
+## Connect to MongoDB Data Sources 
+
+- CRUD & Aggregation APIs: Add `$setisSubset` to list of unsupported pipeline operators.
+
+## Reference
+
+- App Services CLI: Update CLI reference with the `--use-local-app` flag addition to the `functions run` command, the `--limit` flag addition to the `users list` commands, and the new `functions list` command.
```
dacharyc, line 17: Just one minor comment - if we're following the same pattern here as in the rest of the docs, we should probably break down these changes into sub-bullets based on each command - i.e.:

> - App Services CLI
>   - `functions run`: Add `--use-local-app` flag
>   - `users list`: Add `--limit` flag
>   - `functions list`: New command to list all Functions in an app (or whatever this command does)

dacharyc (approved):
Overall, LGTM - just one non-blocking comment, up to you.

PR #779, opened 2024-06-24
Fix incorrect Swift version number for coll-in-mixed feature

cbullinger:
Quick fix: 
Correct version number for Collections in Mixed feature release (should be `10.51.0` instead of `10.15.0`)

- [Data Model Mapping](https://preview-mongodbcbullinger.gatsbyjs.io/atlas-app-services/no-jira-fix-version-number/sync/data-model/data-model-map/#collections-in-mixed-properties)

### Release Notes

**Device Sync**
- Configure and Update Your Data Model > Data Model Mapping: Correct the minimum Swift SDK version number listed in the Collections in Mixed note.


dacharyc (approved):
LGTM - thanks for fixing this up!

PR #796, opened 2024-07-10
DOCSP-35175 Refresh permissions when custom user data changes

lindseymoore:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-35175

- [Define Custom User Metadata](https://preview-mongodblindseymoore.gatsbyjs.io/atlas-app-services/DOCSP-35175/users/custom-metadata/#modify-user-permissions-in-custom-user-data-documents): Add new section on editing permissions in custom user data documents

### Reminder Checklist

- [ ] Did you tag pages appropriately?
  - genre
  - programming_language
  - meta.keywords
  - meta.description

### Release Notes

Authenticate and Manage Users
- Define User Metadata: Create a new section on updating permissions using custom user data.

### Review Guidelines

### DIFF
--- a/source/users/custom-metadata.txt
+++ b/source/users/custom-metadata.txt
@@ -107,6 +107,39 @@ collection. Depending on your use case, you may:
   standard CRUD operations from a Function, an Atlas Device SDK, a MongoDB
   driver, or MongoDB Compass.
 
+Modify User Permissions in Custom User Data Documents
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Permissions will automaticlly refresh for a user when you change their custom 
+user data document. Their :ref:`user session <user-sessions>` 
+will terminate and then reconnect automatically. For user permissions to refresh
+automatically, the custom user data documents should be stored in a normal collection (ref),
+not a view (ref) or timeseries collection (ref). 
+
+For permissions to refresh automatically, keep in mind the following when 
+removing permsisions from a user:
+ 
+1. Don't delete a custom user data document. Rather, unset the fields in the document. 
+
+   .. example::
+
+     Say you have the following custom user data document:
+
+     .. code-block:: json
+        :caption: Need a custom user data document example.???????
+
+        {
+          "_id": "63ed2e4fb7f367c92578e526",
+          "user_id": "63ed2dbe5960df2af7fd216e",
+          "preferences": {
+            "preferDarkMode": true
+          },
+          "dateOfBirth": "1989-03-11T00:00:00.000Z"
+        }
+        
+
+2. Don't unset the user ID field. Rather, unset all the non-ID fields.
+
 Secure Custom User Data
 ~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/source/users/custom-metadata.txt
+++ b/source/users/custom-metadata.txt
@@ -123,36 +123,35 @@ Rather, unset all the non-ID fields in the document.
 
 .. example::
 
-  Consider a collection named ``employees`` where each employee has their own 
-  document stating their level at the company. This collection might have 
-  three roles: Junior, Senior, or Manager. 
-  
-  The ``level`` field in the employee's custom user data document also has 
-  Junior, Senior, or Manager values. Using the ``apply_when`` expression, the 
-  role for the employee is based on the value of the ``level`` field. For example, 
-  the ``apply_when`` expression for the role of Senior would look like:
-
-  .. code-block:: json
-    :caption: Senior role with apply_when expression
-
-    {
-      "name": "Senior"
-      "apply_when": {"%%user.custom_data.level": "senior"},
-      ...
-    }
-
-  Say you you would like to remove the following employee's document
-  because they no longer work at the company:
+  Consider the following document where the user is assigned read and write
+  permissions: 
 
   .. code-block:: json
     :caption: Custom user data document
 
     {
+
       "_id": "63ed2erealobjectid78e526",
       "user_id": "63ed2dbe5960df2af7fd216e",
-      "level": "senior"
+      "canRead": true,
+      "canWrite": true,
     }
-  
+
+  The ``canRead`` and ``canWrite`` fields help determine the roles for the collection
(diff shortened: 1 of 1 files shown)


source/users/custom-metadata.txt
```diff
@@ -107,6 +107,67 @@ collection. Depending on your use case, you may:
   standard CRUD operations from a Function, an Atlas Device SDK, a MongoDB
   driver, or MongoDB Compass.
 
+Modify User Permissions in Custom User Data Documents
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:ref:`Permissions <permissions>` will automaticlly refresh for a user when you
```
samuel-ping, line 113: typo on automatically

source/users/custom-metadata.txt
```diff
@@ -107,6 +107,67 @@ collection. Depending on your use case, you may:
...
+
+  .. code-block:: json
+    :caption: Correctly updated custom user data document
+
+    {
+      "_id": "63ed2erealobjectid78e526",
+      "user_id": "63ed2dbe5960df2af7fd216e"
+    } 
+
+  Unsetting the non-ID field allows App Services to remove the employee's
```
samuel-ping, line 167: There's a slight detail where if you delete the document instead of unsetting the field, we just won't _automatically_ update the employee's permissions. The permissions would still be updated if the end user closes and re-opens their app, for example.

Also, deleting the custom user data document doesn't necessarily mean that the employee's permissions are removed, as they could still have some permissions if the developer has a role set up with an empty `apply_when` expression as a catch-all.

source/users/custom-metadata.txt
```diff
@@ -107,6 +107,67 @@ collection. Depending on your use case, you may:
   standard CRUD operations from a Function, an Atlas Device SDK, a MongoDB
   driver, or MongoDB Compass.
 
+Modify User Permissions in Custom User Data Documents
```
samuel-ping, line 110: I think a more apt title might be along the lines of "Best Custom User Data Practices for Automatic Refreshing Permissions"

Reason being, I feel like the current title gives the idea that custom user data documents alone can set permissions, when in reality its all dependent on the roles the developer has set up.

samuel-ping (changes requested):
This is looking really good so far! I think I wasn't completely clear when explaining some things to you, that's definitely on me- the comments I left should address those.

I think an overarching idea is that on their own, custom user data documents don't do anything. They are just extra fields of information you can store for a user. You can reference custom user data in your roles, and the way your roles are set up will determine the permissions for a user.

Let me know if you have any more questions about anything!

samuel-ping (commented):
Also I know I provided you with the current example, but I actually think it would make sense to simplify the example to just the example provided in the Jira ticket. So for example with the following custom user data document:
```
{
   "_id": "63ed2erealobjectid78e526",
    "user_id": "63ed2dbe5960df2af7fd216e",
    "canRead": true,
    "canWrite": true,
}
```
If you wanted to remove this user's document and have the resulting permissions be automatically reflected for the user's sessions, you should first unset all of their non-ID fields. After that change, the above document would become:
```
{
   "_id": "63ed2erealobjectid78e526",
    "user_id": "63ed2dbe5960df2af7fd216e",
}
```
and the user's permissions will automatically refresh according to the roles. After that you can safely delete the custom user data document if you wish.

source/users/custom-metadata.txt
lindseymoore: Simplified to Permissions and Custom User Data. I can workshop the title with copy reviewer, thanks!

source/users/custom-metadata.txt
```diff
@@ -107,6 +107,66 @@ collection. Depending on your use case, you may:
   standard CRUD operations from a Function, an Atlas Device SDK, a MongoDB
   driver, or MongoDB Compass.
 
+Permissions and Custom User Data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:ref:`Permissions <permissions>` will automatically refresh for a user when you 
+change their custom user data document. Their :ref:`user session <user-sessions>` 
+will terminate and then reconnect automatically.
```
samuel-ping, line 115: I think the most apt term to use here would be "refresh" for user sessions.
```suggestion
will refresh automatically. 
```

source/users/custom-metadata.txt
```diff
@@ -107,6 +107,66 @@ collection. Depending on your use case, you may:
...
+.. example::
+
+  Consider the following document where the user is assigned read and write
+  permissions: 
+
+  .. code-block:: json
+    :caption: Custom user data document
+
+    {
+
```
samuel-ping, line 133:
```suggestion
```

source/users/custom-metadata.txt
```diff
@@ -107,6 +107,66 @@ collection. Depending on your use case, you may:
...
+      "_id": "63ed2erealobjectid78e526",
+      "user_id": "63ed2dbe5960df2af7fd216e",
+      "canRead": true,
+      "canWrite": true,
+    }
+
+  The ``canRead`` and ``canWrite`` fields help determine the roles for this document's
```
samuel-ping, line 140: Maybe add the word "can", since the fields don't necessarily always help unless the roles reference them
```suggestion
  The ``canRead`` and ``canWrite`` fields can help determine the roles for this document's
```

samuel-ping (approved):
Everything looks good technically! I just left a couple of super nitpicky optional nitpicks, but I'll leave it up to you to decide whether you want to make the changes.

source/users/custom-metadata.txt
```diff
@@ -107,6 +107,65 @@ collection. Depending on your use case, you may:
   standard CRUD operations from a Function, an Atlas Device SDK, a MongoDB
   driver, or MongoDB Compass.
 
+Permissions and Custom User Data
```
cbullinger, line 110:
```suggestion
Refresh User Permissions for Custom User Data
```
can't think of a better wording, but this is more about refreshing permissions, right?

cbullinger (approved):
one suggestion, but otherwise lgtm!

PR #806, opened 2024-07-12
Adds github action to generate preview links in PR description

MongoCaleb:
## Pull Request Info

Adds github action to generate preview links in PR description. Updated the template to provide a placeholder for regex to match again (and know where to insert links).
This Action does **not** build links for files in /source/examples, /source/includes, or /source/images.

### Future work:
- If there is an include, provide links to pages that contain that include
- same with an image

 - No pages to preview

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,7 +2,8 @@
 
 Jira ticket: https://jira.mongodb.org/browse/DOCSP-NNNNN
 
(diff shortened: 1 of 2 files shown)


dacharyc (changes requested):
This seems like a reasonable approach, but I believe we need to separate this into two workflows for security reasons. I've linked out to where Chris did this work in the Readability workflow, so hopefully it can help you figure out the right approach.

dacharyc (approved):
LGTM - this should be a nice improvement, thanks for putting this together! 🚀

PR #856, opened 2024-09-25
autobuilder check is no longer needed with Netlify builds

MongoCaleb:
The autobuilder check is no longer needed with Netlify builds.


dacharyc (approved):
Nice! :shipit:

PR #880, opened 2025-04-08
Data API migration guide: Clarify API key and fix broken Node.js link

cbullinger:
## Pull Request Info

No Jira ticket; reported via Slack 

- Correct and clarify that the API key/secret used is only a locally defined key for basic auth, and is not an Atlas API key
- Fix broken link to node.js docs


krollins-mdb (approved):
LGTM! Thanks for the quick updates, @cbullinger.

